PR #35, opened 2014-09-10
Deprecation warnings from psutil

datasaur:
psutil 2.1.1 is warning about deprecated functions get_memory_percent() and get_cpu_percent() in the header of the UI. This simply calls the functions as suggested by the warning.


julmon:
Thanks for it, but already fixed in a more global way.

PR #45, opened 2015-10-14
Adds support for using pg_activity with Amazon RDS

NickLaMuro:
Taken from the following blog post and fork:

http://blog.geesu.net/2015/06/09/using-pg_activity-with-amazon-rds/
https://github.com/Geesu/pg_activity/tree/feature/amazon-rds

Just adds this as an command line option instead of forcing it on all users.  Should maintain backwards compatibility with the previous functionality, just allows this to be used without errors when running against a remote RDS database.  All credit should go to @Geesu as he was the person that did the original legwork.

I can add documentation if you want, but I kept it minimal for now incase you would prefer a different approach to solving this problem.

### DIFF
--- a/pg_activity
+++ b/pg_activity
@@ -119,6 +119,13 @@ server activity monitoring.")
                 help = "Filesystem blocksize (default: 4096)",
                 metavar = 'BLOCKSIZE',
                 default = 4096)
+            # --rds
+            parser.add_option(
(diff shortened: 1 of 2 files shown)


Geesu:
:+1:

julmon:
Thank you!

PR #46, opened 2015-10-22
Require python2. Fixes syntax errors on OS-es where "python" is python3.

patryk-s:
I'm using Arch Linux and the default python environment is python3. When trying to run pg_activity I get:

    File "./pg_activity", line 211
      except OptionParsingError, err:
                               ^
    SyntaxError: invalid syntax

### DIFF
--- a/pg_activity
+++ b/pg_activity
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 """
 pg_activity


julmon:
Hi,

I think adding python3 support should be a better way to handle that case. Just commited some changes to support python3: https://github.com/julmon/pg_activity/commit/6a65974dcc23e84aee36f2ec221711af365ee2ed

Could you please try with this commit if pg_activity is running well in your environment ?

Thank you,

patryk-s:
Sorry it took so long to answer but it works now on python3. Thanks.

PR #47, opened 2015-12-02
Manpage fixes

k0lter:
Update manpage's NAME section so that tools such as whatis(1) or apropos(1) could use it


julmon:
Thank you :+1:

Issue #51, opened 2016-08-12
Chooses wrong password when connecting


feikesteenbergen:
When trying to connect to a localhost connection, the wrong password is supplied, this is my `.pgpass`:

```text
localhost:*:*:fsteenbergen:password1
*:*:*:postgres:postgres
*:*:*:fsteenbergen:password2
```

It takes the password `password2`

As psycopg2 will take care of correct `PGPASSFILE` handling itself, why not just drop the magic?

PR #52, opened 2016-08-12
Depend on psycopg2 behaviour for passwords

feikesteenbergen:
Don't do anything to get the password from the PGPASSFILE.
Psycopg2 uses libpq which does all the password handling correctly.

Addresses issue #51


julmon:
Thanks !

PR #57, opened 2016-10-22
add tests, move option parser to separate module

fried-sausage:
Added tests.
As of now they only check 2 methods and 1 function from UI module.
I`d appreciate ideas regarding what else and how should be tested.

Added package version and name to `__init__.py`
It is customary to put methainformation like this to `__init__`.
You can now import application version from anywhere in package.

Moved parser from main script to separate module.
If there is need to bring changes to argument parser in future (for example implement it with argparse instead of deprecated optparse, or add some options), we would only need to edit designated module. We also reduce amount of code in main script.


julmon:
Hi,

Many thanks for this huge amount of work!

When I try to run `pg_activity`, I got this error:
```
  File "./pg_activity", line 44, in <module>
    PGAUI = UI.UI()
TypeError: __init__() takes exactly 2 arguments (1 given)
```

fried-sausage:
@julmon 
Sorry, im not done with this yet.
I will fix this as soon as i have some time and ping you.

fried-sausage:
Sorry, i abandoned this project completely, guess it is too late to fix this now.
It would be more honest if i just close this PR.

PR #65, opened 2017-06-06
Issue 64 add support to version 10

fabriziomello:
Support to new version numbering strategy (X.Y - Major.Patch) starting for upcoming 10 version.


julmon:
Thanks !

PR #70, opened 2017-10-31
process['database'] can be None for some maintenance processes

olasd:
We use barman with wal shipping for PITR on our main database. This creates a
walsender process that is not connected to any specific database.


julmon:
Thanks !

PR #90, opened 2018-10-02
Doc update about system info + examples

Krysztophe:
Added a sentence in the man page for people like me that didn't read the README on github and didn't know that pg_activity could show system information, provided it runs as the right user.

Added an example, and updated the one with `-h /tmp` (not used anymore on Debian and RH since 9.4 at least).

I suppose that I had to run build-man.sh.


julmon:
Thank you !

PR #95, opened 2018-12-12
[IMP] new option to avoid total db sizes

nseinlet:
On clusters with thousands of databases, computing
total size of the cluster is painfull and make
pg_activity unusable. Don't compute the value make
pg_activity usable on those large clusters.


nseinlet:
@julmon Can you review it to make pg_activity usable on clusters with many databases?

julmon:
Thank you !

PR #99, opened 2018-12-12
Add --output option

### LINKED ISSUE #62: Should have option to store RUNNING QUERIES states
To store running queries value into log file using command line option should be there.
If pg_activity cannot extend the feature then, can you tell me which function is responsible for building data for a terminal.
So I can compare  my log to analyse performance.

julmon:
This new option defines a file path where to store running
queries/backends using CSV format. Default value is None, meaning the
feature is not active.
Running queries storage is working when pg_activity is on RUNNING
QUERIES tab only.

Fix #62

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Options
 		-C, --no-color        Disable color usage.
 		--blocksize=BLOCKSIZE Filesystem blocksize (default: 4096).
 		--rds                 Enable support for AWS RDS.
+		--output=FILEPATH     Store running queries as CSV.
 		--help                Show this help message and exit.
 		--debug               Enable debug mode for traceback tracking.
 		--no-db-size          Skip total size of DB.
--- a/README.md
+++ b/README.md
@@ -71,6 +71,13 @@ Options
     	--no-wait             Disable W.
     	--no-app-name         Disable App.
 
+
+Notes
+-----
+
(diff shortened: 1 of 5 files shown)


michelmilezzi:
Awesome! 
Just small suggestions:

`pg_stat_activity.query` relies on `track_activity_query_size`. Perhaps it's a good idea adding a small note to readme in the case the user gets your queries truncated.

It's not possible to use `COPY ... WITH CSV` instead of [2192](https://github.com/julmon/pg_activity/pull/99/files#diff-7e2823caf56657efe3c04aa67a6ea538R2192)?

Thanks!

julmon:
> `pg_stat_activity.query` relies on `track_activity_query_size`. Perhaps it's a good idea adding a small note to readme in the case the user gets their queries truncated.

Good catch. Just added a note about this.

> It's not possible to use `COPY ...TO STDOUT WITH CSV` instead of [2192](https://github.com/julmon/pg_activity/pull/99/files#diff-7e2823caf56657efe3c04aa67a6ea538R2192)?

Actually, we have to do some intermediate calculations between activity reading from `pg_stat_activity` view and displaying the output (system metrics computation), I don't think that working with CSV data source at this level would help.

ahes:
Output saving is a nice option but how to quit pg_activity after specified time, for example 5 seconds?

PR #114, opened 2019-05-14
Adding refresh dbsize interative action

fabiogeiss:
764/5000
Hello Lulien. I really loved the option --no-dbsize. It's something I was already thinking of implementing because I was having load problems due to the high cost of the query. Since its availability, I have always used it, but many times I have wanted to see what the current size of the base is. I needed to run another instance of pg_activity without the --no-dbsize option only to see the size of the base and just after closing it. I then created an interactive option for this, using the "D" to signal the refresh. I'm not a python programmer, so I don't know if I did it the best way. Feel free to perform any refactorings you deem appropriate. Thank you in advance for accepting my pull request. It's really very enjoyable for me to contribute to this very cool project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 pgactivity/__pycache__/
+.vscode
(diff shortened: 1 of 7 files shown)


julmon:
Thank you !

PR #116, opened 2019-07-18
Expand current 1-3s refresh interval to 0.5-5s

kmoppel:
Increment size decreases to 0.5s when going below 1s


kmoppel:
Sometimes I personally find the current 1s not enough "real time" and sometimes on a very loaded system would do good also to have a bit bigger interval.

julmon:
Thank you !

Issue #118, opened 2019-08-07
Make pip installs easier by depending on psycopg2-binary vs psycopg2


kmoppel:
When not using native package managers like APT but "pip" directly then the "psycopg2" package requirement means some other extra packages are needed. psycopg2-binary is API compatible with psycopg2 and has no extra dependencies.

There might be some caveats though, so would make sense to read this discussion - https://github.com/psycopg/psycopg2/issues/674

blogh:
Hi,

Thank you for reporting this.
We have decided to use the same approach as [patroni](https://github.com/zalando/patroni/pull/102) and [temboard](https://github.com/dalibo/temboard/commit/2ffa8777123504e8c2d01fb16fb058e6d37c4ad0). We will remove psycopg2 from setup.py and document that the rpm should be installed beforehand.

Thx
Benoit

blogh:
Done as stated above.

PR #126, opened 2020-01-30
Doc update and setup.py adjustment for psycopg2

blogh:
see #118

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Dependencies
 Installation from sources:
 `setuptools` &ge; **0.6.14**
 
-In order to use `pg_activty` after an install using `pip`, you have to install the package `psycopg2` from `pgdg` first (see [here](https://github.com/dalibo/pg_activity/issues/118) for more info).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -14,6 +14,8 @@ Dependencies
 Installation from sources:
 `setuptools` &ge; **0.6.14**
 
+In order to use `pg_activty` after an install using `pip`, you have to install the package `psycopg2` from `pgdg` first (see [here](https://github.com/dalibo/pg_activity/issues/118) for more info).
```
julmon: I think we could reword it a bit like that:

>  you have to install the package `psycopg2`  from `pgdg` [APT](https://wiki.postgresql.org/wiki/Apt) or [YUM](https://yum.postgresql.org/) repositories. `psycopg2` can also been installed from `pip` with `pip install psycopg2` or `pip install psycopg2-binary` for the binary version.

blogh: thx, works for me. fixed that way in [50ffa06](https://github.com/dalibo/pg_activity/pull/126/commits/50ffa06bd9d930fac85159ee79ac58154c759af9)

PR #129, opened 2020-02-15
Introduce an explicit "duration mode"

nilshamerlinck:
- so far the content of the TIME column only reflected the duration of the query
- this PR introduces 2 other duration modes: transaction, backend
- press T to cycle through different duration modes
- active duration mode is shown in the top status bar:

duration_mode


nilshamerlinck:
Done @julmon

julmon:
Thanks !

Issue #130, opened 2020-02-17
'utf-8' codec can't decode byte 0xf8 in position 0


debnet:
Since the installation of `powa`, I got error with `pg_activity` script whereas I'm using it for years without a single error.

```
(python) [marc@marcbook ~]$ pg_activity -U postgres
FATAL: 'utf-8' codec can't decode byte 0xf8 in position 0: invalid start byte
```

julmon:
Could you try with `--debug` ?

debnet:
`Traceback (most recent call last):
  File "/usr/bin/pg_activity", line 310, in main
    procs, disp_procs)
  File "/usr/lib/python3.7/site-packages/pgactivity/UI.py", line 1166, in poll
    disp_proc)
  File "/usr/lib/python3.7/site-packages/pgactivity/UI.py", line 1295, in __poll_activities
    queries =  self.data.pg_get_activities()
  File "/usr/lib/python3.7/site-packages/pgactivity/Data.py", line 448, in pg_get_activities
    ret = cur.fetchall()
  File "/usr/lib64/python3.7/site-packages/psycopg2/extras.py", line 99, in fetchall
    res = super(DictCursorBase, self).fetchall()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf8 in position 0: invalid start byte
FATAL: 'utf-8' codec can't decode byte 0xf8 in position 0: invalid start byte`

Seems the error come from `psycopg2` but I was using `pg_activity` for years without error, I got this error since `powa` is installed on my local database.

blogh:
Hi, 

The problem is related to this issue : https://github.com/powa-team/powa-archivist/pull/20
The fix is not in the current stable version of  PoWA.

The backend_type is wrong :
<pre>
10.20.40.50:5432 powa@postgres=# SELECT application_name, backend_type FROM pg_stat_activity \gx
-[ RECORD 1 ]----+--------------------------------
application_name | PoWA collector
backend_type     | pm\x7F
</pre>

Which triggers :
<pre>
UnicodeDecodeError('utf-8',  b'pm\xa7\x95\xff\x7f', 2, 3, 'invalid start byte')
</pre>

The PR attempts to fix this.

Benoit.

blogh:
The PR was merged, can you try it ?

blogh:
Fixed

PR #134, opened 2020-04-09
Replace backend_type with is_background_worker

blogh:
* This commit attempts to solve the issue : #130 [1].
* PoWA has a bug where `backend_type` is not correctly filled [2]. it
  was fixed but it's not in the stable version yet.
* Since [3], we queried for the `backend_type` and filtered
  `parallel workers` in the Python code. The test is now done in
  SQL,  which avoids the problems all together.
* The `parallel worker` detection was not working for pg 11+ because
  it uses the 'parallel worker' backend_type. It's now fixed.

[1] https://github.com/dalibo/pg_activity/issues/130
[2] https://github.com/powa-team/powa-archivist/pull/20
[3] https://github.com/dalibo/pg_activity/commit/6a180d86139d68a728946997996f346cc880888e

### DIFF
--- a/pgactivity/UI.py
+++ b/pgactivity/UI.py
@@ -2235,10 +2235,9 @@ class UI:
         dif = self.maxx - len(indent) - 1
 
         query = ''
-        if 'is_parallel_worker' in process and \
-                process['is_parallel_worker']:
-
+        if process.get('is_parallel_worker'):
             query += '\_ '
+
         if self.verbose_mode == PGTOP_TRUNCATE:
             query += process['query'][:dif]
             colno += self.__print_string(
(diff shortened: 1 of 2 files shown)


pgactivity/UI.py
```diff
@@ -2235,8 +2235,8 @@ def __refresh_line(self, process, flag, indent, \
         dif = self.maxx - len(indent) - 1
 
         query = ''
-        if 'backend_type' in process and \
-                process['backend_type'] == 'background worker':
+        if 'is_parallel_worker' in process and \
```
julmon, line 2238: could be simply done with:
```python
if process.get('is_parallel_worker'):
```

blogh: done that way.

PR #144, opened 2020-07-07
some cleanups while reading the code

dlax:
A couple of cleanup changes I found useful while reading through the code (especially the main and UI files).

First commits contain refactorings and fixes about the argument parser.
Then, come some style changes to make flake8 less noisy (to a point I find it usable in my editor).


dlax:
Will mostly be included in #145

PR #145, opened 2020-07-16
UI rewrite

dlax:
This PR concerns the rewrite of the UI part of `pg_activity` using [blessed](https://pypi.org/project/blessed/). It also includes packaging improvements and introduces a modern QA setup.


dlax:
This PR is mostly done, the UI has been totally rewritten while preserving the actual behavior as much as possible. I don't expect anybody to review these 140+ commits (though, this is welcome :)), but would very much appreciate any feedback on the functional side.

To test this, pull this branch, `pip install` the project in a Python 3 virtualenv and run `pg_activity` to get the new UI, to be compared with the old one still available as `pg_activity --legacy-ui`. Unless I missed something (which is quite possible), all options are supported except `--no-color`, which has not (yet?) been ported in the new UI. The new code is quite well tested, with currently 97% of code coverage concerning rewritten modules.

The next step is to remove the old code, this is also mostly done as followed, see  https://github.com/dlax/pg_activity/compare/ui-rewrite...drop-old-ui

CC @blogh and @julmon

dlax:
I actually missed the "navigation mode"; will work on this later on.

dlax:
I've now implemented the interactive mode and I think this is (almost) ready. The old UI code got dropped and many things cleaned up.

@blogh can you test again?

Next I'd like to merge this PR and make a pre-release (maybe 2.0a1) and request further feedback.

dlax:
Last commits add support for the "tag" action in interactive mode.

julmon:
Huge work @dlax! I will give it a try next week. Many thanks!

PR #150, opened 2020-11-02
Anosql

blogh:
add https://github.com/honza/anosql
should probably change to https://github.com/nackjicholson/aiosql later on

### DIFF
--- a/pgactivity/Data.py
+++ b/pgactivity/Data.py
@@ -283,7 +283,7 @@ class Data:
         skip_db_size = skip_sizes and (not self.refresh_dbsize)
 
         ret = self.queries.get_db_info(
-            self.pg_conn, 
+            self.pg_conn,
             using_rds=using_rds,
             skip_db_size=skip_db_size,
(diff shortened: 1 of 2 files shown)


blogh:
Some queries have to be retrieved with anosql and executed normally because some parameters replace column names which is not yet supported by either anosql or aiosql. There is an issue about this in aiosql (https://github.com/nackjicholson/aiosql/issues/21)

blogh:
@dlax : if feels weird to have queries executed thru `anosql` and others via `psycopg2`.
Do you think I should only use `anosql` to store the queries and execute tem all with `psycopg2` ?

dlax:
That would be a reasonable compromise and a good first step.

On the other hand, if we cannot fully use anosql now because of this limitation, I wonder if we shouldn't instead roll out our own query-loader (a function returning python strings from a SQL file should be fairly easy to implement). That would also avoid pulling a dependency that is not packaged in CentOS.

PR #156, opened 2020-11-27
Fix dbname defaults

blogh:
dbname default values should be :
* PGDATABASE ;
* or a database whith the same name as the username of the connection.

### DIFF
--- a/pg_activity
+++ b/pg_activity
@@ -86,9 +86,9 @@ server activity monitoring.")
         '-d',
         '--dbname',
         dest = 'dbname',
-        help = "Database name to connect to (default: \"postgres\").",
+        help = "Database name to connect to (default: same as user).",
         metavar = 'DBNAME',
-        default = 'postgres')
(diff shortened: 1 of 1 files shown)


dlax (approved):

blogh:
After looking back into this PR, event with this, it will still not behave like `psql`.

The problem is that it should not be `pg_activity`'s job to handle this logic, it's already done by the client library.

> Any other connection parameter supported by the client library/server can be passed either in the connection string or as a keyword. The PostgreSQL documentation contains the complete list of the supported parameters. Also note that the same parameters can be passed to the client library using environment variables.
Link : [psychopg2.connect](https://www.psycopg.org/docs/module.html?highlight=connect#psycopg2.connect)

It's easy to see by commenting all default in the parameters.

If we remove the default values, it breaks the display of the connection url because we display what was provided to `psycopg2.connect()` by us and not what was used by the client library (which is provided by `psycopg2.extensions.ConnectionInfo() = connection.info`). 

I am for dropping this PR and maybe revamp all the connection thing later which would make it simplier (I think).

Note : I would require to test will old versions.

PR #154, opened 2020-11-26
Connection string

blogh:
Add connection strings to address https://github.com/dalibo/pg_activity/issues/151 & https://github.com/dalibo/pg_activity/issues/147

The idea is to mimic psql behavior of allowing a connection string to be specified as argument of the script. In that case the information provided overrides the host, port, user and dbname options.

Two syntax are possible  (depending in the version of libpq). (https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING)

```
pg_activity "host=/tmp port=5432 user=toto dbname=pgbench"
pg_activity "postgresql://toto@localhost:5432/bench"
```
I am not very fond of the argument thing, I would have preferred an option like --dsn or somehing. It's less ugly in my pov but it's done like that in psql.
 
Note: I also fixed the default value of dbname which was wrong.

PR #157, opened 2020-11-27
refactoring to prepare support for dsn options

dlax:


dlax:
These changes will be part of #154, closing.

PR #162, opened 2021-01-05
Use conninfo to populate the Host value

dlax:
Follow-up on https://github.com/dalibo/pg_activity/pull/156#issuecomment-754658744, @blogh. 

The UI behaves slightly differently as it would now display the socket value (e.g. `/var/run/postgresql`) instead of `localhost` in the header line for local connection.

### DIFF
--- a/pg_activity
+++ b/pg_activity
@@ -33,8 +33,8 @@ import getpass
 from optparse import OptionParser, OptionGroup
 from typing import NoReturn
 
-from blessed import Terminal
 import psycopg2
+from blessed import Terminal
 
 from pgactivity import __version__, data, types, ui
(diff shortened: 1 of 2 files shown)


blogh:
`/var/run/postgresql` or `/tmp` are the default values depending on the platform.

>    The default behavior when host is not specified, or is empty, is to connect to a Unix-domain socket in /tmp (or whatever socket directory was specified when PostgreSQL was built). On Windows and on machines without Unix-domain sockets, the default is to connect to localhost.
> link  : https://www.postgresql.org/docs/13/libpq-connect.html#LIBPQ-CONNECT-HOST

PR #163, opened 2021-01-14
Handle disconnection from postgres and try to reconnect

dlax:
We restore the previous behavior (sort of) that got dropped during ui rewrite by trying 3 times to reconnect in case of disconnection, waiting 5s between each attempts.


dlax:
cc @blogh, if you want to try this out.

blogh:
cool, I completely missed it while testing.

blogh:
works great.

blogh:
thanks

PR #173, opened 2021-02-03
Some cleanups and improvements to SQL queries

dlax:
Follow-up on #171, cleaning some things while reading queries. (Read each commits for details.)


blogh:
looks good to me.

PR #179, opened 2021-02-17
Cleanup defaults

blogh:
Remove all the default values set in pg_activity to let libpq do it's thing.

### DIFF
--- a/pg_activity
+++ b/pg_activity
@@ -79,8 +79,7 @@ def get_parser() -> ModifiedOptionParser:
         "-U",
         "--username",
         dest="username",
-        default=os.environ.get("PGUSER") or getpass.getuser(),
-        help="Database user name (default: %default)",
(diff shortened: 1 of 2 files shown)


blogh:
@dlax : It's seems that since the change in exception handling in pg_connect incorrect password is no longer trapped. Am doing something wrong ?

Test on master :
```
(pgactivity) [benoit@benoit-dalibo pg_activity]$ pg_activity "host=/tmp port=5436 user=toto dbname=postgres"
pg_activity: FATAL: fe_sendauth: no password supplied
```

dlax (approved):

dlax:
> It's seems that since the change in exception handling in pg_connect incorrect password is no longer trapped. Am I doing something wrong ?

Confirmed. We have dropped something by mistake.

Issue #158, opened 2020-12-01
PostgreSQL Log related enhancements


blogh:
* Add an option to hide pg_activity's query in the logs when possible
* Add an appname

PR #183, opened 2021-02-26
Add appname and control logs (WIP)

blogh:
In response to : https://github.com/dalibo/pg_activity/issues/158
I have added application name and an option to disable query logging for pg_activity.

### DIFF
--- a/pg_activity
+++ b/pg_activity
@@ -264,7 +264,7 @@ def get_parser() -> ModifiedOptionParser:
     # --hide-queries-in-logs
     group.add_option(
         "--hide-queries-in-logs",
-        dest="hidequeriesinlogs",
+        dest="hide_queries_in_logs",
         action="store_true",
         help="Disable log_min_duration_statements and log_min_duration_sample for pg_activity",
         default=False,
--- a/pg_activity
+++ b/pg_activity
@@ -261,6 +261,14 @@ def get_parser() -> ModifiedOptionParser:
         help="Disable App",
         default=False,
     )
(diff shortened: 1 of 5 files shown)


dlax:
About application_name, I think we can define this when connecting with psycopg: `psycopg2.connect(..., application_name="pg_activity")`.

blogh:
I was wondering if the approach using queries is correct so I tried using dsn's like :

```
application_name=pg_activity options='-c log_min_duration_statement=-1'
```

The problem being that some parameters are version dependant which requires a connection to be aware of.

blogh:
Just read your message, will try it your way.

blogh:
Indeed ... Better ...

pg_activity
```diff
@@ -261,6 +261,14 @@ def get_parser() -> ModifiedOptionParser:
         help="Disable App",
         default=False,
     )
+    # --hide-queries-in-logs
+    group.add_option(
+        "--hide-queries-in-logs",
+        dest="hidequeriesinlogs",
```
dlax, line 267: This would be more readable as `hide_queries_in_logs`.

dlax (approved):
LGTM, modulo the nit about option name.

When merging, I'd suggest to squash commits to have a clean history and write a meaningful commit message. Thanks.

pg_activity
blogh: I agree, I used this because all the other option names where squashed like this.
Will used your format, then.

blogh: done

PR #187, opened 2021-03-04
Try to reconnect indefinitely

### LINKED ISSUE #184: Reconnect when the server restarts
When a server restarts, pg_activity fails.

I'd like to see reconnect itself alone by default
( provided there is no option like eg  ''--timeout'' ou ''--no-retry'', but I'm not sure they would be so useful).

dlax:
Though catch CTRL+C to abort the loop.

Fix #184


blogh:
LGTM

PR #188, opened 2021-03-04
Include queries data in sdist and installation

dlax:
We follow recommendations from:

  https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html


blogh:
works for me

PR #200, opened 2021-03-17
Fix version decoding for pg >= 10

blogh:
* Fix version decoding for version >= 10 as reported in issue 199
* Fix a bug with devel version < 10
* Fix tests

### DIFF
--- a/pgactivity/data.py
+++ b/pgactivity/data.py
@@ -29,19 +29,21 @@ def pg_get_num_version(text_version: str) -> Tuple[str, int]:
     version()).
 
     >>> pg_get_num_version('PostgreSQL 11.9')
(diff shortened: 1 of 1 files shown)


dlax (approved):
LGTM.

Aside, I think we should drop most of this code and rely on psycopg/libpq to retrieve the server version (integer): https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.ConnectionInfo.server_version

blogh:
Sure, It completely slipped my mind but we already use it [here](https://github.com/dalibo/pg_activity/blob/aeed63ead6bd06c84e54161e71980fb3a448faed/pgactivity/data.py#L134).

Will look into it ASAP.

dlax (approved):

PR #204, opened 2021-03-26
Handle 'pg_stat_activity.datname' column possibly being NULL

### LINKED ISSUE #203: Problem to run pg_activity when there's a barman connection
We have a CentOS 7 server running PostgreSQL version 12 and we use Barman (https://www.pgbarman.org/) as one of our backup tools.

When we try to run pg_activity without any additional parameter, we get the following error:

```bash
[root@server ~]# sudo -u postgres pg_activity -U postgres
  File "/bin/pg_activity", line 11, in <module>
    load_entry_point('pg-activity==2.1.3', 'console_scripts', 'pg_activity')()
  File "/usr/lib/python3.6/site-packages/pgactivity/cli.py", line 293, in main
    ui.main(term, dataobj, host, options, dsn)
  File "/usr/lib/python3.6/site-packages/pgactivity/ui.py", line 227, in main
    width=width,
  File "/usr/lib/python3.6/site-packages/pgactivity/views.py", line 526, in screen
    width=width,
  File "/usr/lib/python3.6/site-packages/pgactivity/views.py", line 123, in wrapper
    for line in func(term, *args, **kwargs):
  File "/usr/lib/python3.6/site-packages/pgactivity/views.py", line 378, in processes_rows
    cell(getattr(process, field), column)
  File "/usr/lib/python3.6/site-packages/pgactivity/views.py", line 363, in cell
    text_append(f"{color}{column.render(value)}")
  File "/usr/lib/python3.6/site-packages/pgactivity/types.py", line 239, in render
    return self.template_h % self.transform(value)[: self.max_width]
  File "/usr/lib/python3.6/site-packages/pgactivity/utils.py", line 79, in ellipsis
    lv = len(v)
TypeError: object of type 'NoneType' has no len()
```

To figure out which database name was coming as `None` and making the function `ellipsis` to fail, I edited the file `types.py` to not apply the function `ellipsis` on the database name. Then, pg_activity ran well and showed a barman connection but with the database name being `None` in fact.

image

At this point I don't know where the problem is happening and why that "zombie connection" (with no reads and no writes) shows up without a database name. One possible solution would be to handle the situations when the database name is None. Something like that:

```python
def ellipsis(v: str, width: int) -> str:
    """Shorten a string to specified width with '...' in the middle.
    >>> ellipsis("loooooooooog", 7)
    'lo...og'
    >>> ellipsis("loooooooooog", 6)
    'lo...g'
    >>> ellipsis("short", 10)
    'short'
    """
    if v is None:
        return "Something here"
    lv = len(v)
    if lv <= width:
        return v
    assert width >= 5
    wl = (width - 3) // 2
    return v[: wl + 1] + "..." + v[-wl:]
```

Thanks in advance for any thought or help in how to fix this issue.

dlax:
Fix #203

### DIFF
--- a/pgactivity/types.py
+++ b/pgactivity/types.py
@@ -303,7 +303,7 @@ class UI:
                 name="DATABASE",
                 template_h=f"%-{max_db_length}s ",
                 transform=functools.lru_cache()(
-                    functools.partial(utils.ellipsis, width=16)
+                    lambda v: utils.ellipsis(v, width=16) if v else "",
                 ),
                 sort_key=None,
             )
@@ -683,7 +683,7 @@ def locktype(value: str) -> LockType:
 class BaseProcess:
     pid: int
     application_name: str
-    database: str
+    database: Optional[str]
     user: str
     client: str
     duration: Optional[float]
--- a/pgactivity/utils.py
+++ b/pgactivity/utils.py
@@ -174,7 +174,7 @@ def csv_write(
     ...      'query': 'autovacuum: VACUUM ANALYZE public.pgbench_tellers',
     ...      'duration': 0.348789, 'wait': False,
     ...      'io_wait': False, 'is_parallel_worker': False},
-    ...     {'pid': 25068, 'application_name': 'pgbench', 'database': 'pgbench',
+    ...     {'pid': 25068, 'application_name': 'pgbench', 'database': None,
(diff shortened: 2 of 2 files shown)


sahapasci (approved):

blogh (commented):
works for me.

Issue #209, opened 2021-05-04
Traceback Printed When User Sends Keyboard Interrupt (SIGINT)


vfrazao-ns1:
Great tool! This one thing has been bothering me - When I press ctrl+c the program will exit but also print a traceback - I pressed Ctrl+C with the intent to exit the program so I don't need the traceback information.

Steps to repro:

1. Start pg_activity
2. Run `pkill -2 pg_activity`
3. Observe results:

```
Traceback (most recent call last):
  File "/usr/local/bin/pg_activity", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/dist-packages/pgactivity/cli.py", line 293, in main
    ui.main(term, dataobj, host, options, dsn)
  File "/usr/local/lib/python3.6/dist-packages/pgactivity/ui.py", line 236, in main
    key = term.inkey(timeout=ui.refresh_time) or None
  File "/usr/lib/python3/dist-packages/blessed/terminal.py", line 1126, in inkey
    while not ks and self.kbhit(timeout=_time_left(stime, timeout)):
  File "/usr/lib/python3/dist-packages/blessed/terminal.py", line 931, in kbhit
    ready_r, _, _ = select.select(check_r, [], [], timeout)
```

dlax:
Fixed by #210, thanks again!

PR #210, opened 2021-05-04
Gracefully handle keyboard interupt (SIGINT)

vfrazao-ns1:
Relates to issue: https://github.com/dalibo/pg_activity/issues/209

First time contributor - please let me know if I'm missing anything for this PR (I didn't see any contributing or dev guideline).

### DIFF
--- a/pgactivity/__main__.py
+++ b/pgactivity/__main__.py
@@ -1,10 +1,4 @@
-import sys
-
 from .cli import main
 
 if __name__ == "__main__":
-    try:
-        main()
-    except KeyboardInterrupt:
-        sys.exit(1)
-
+    main()
--- a/pgactivity/__main__.py
+++ b/pgactivity/__main__.py
@@ -1,4 +1,10 @@
+import sys
+
 from .cli import main
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


pgactivity/__main__.py
```diff
@@ -1,4 +1,10 @@
...
+
 from .cli import main
 
 if __name__ == "__main__":
-    main()
+    try:
+        main()
+    except KeyboardInterrupt:
+        sys.exit(1)
+
```
dlax, line 10: This change does seem needed, as the change in `cli.main()`, below is sufficient I think.

(This file is used as `python3 -m pgactivity [ARGS]`.)

dlax (changes requested):
Thank you for your contribution.

It looks good to me overall, just one thing apparently superfluous.

pgactivity/__main__.py
vfrazao-ns1: Yes, thank you for pointing that out. I've changed `__main__.py` back to match master.

dlax (approved):
LGTM. Will merge when CI completes.

Thanks!

PR #211, opened 2021-05-06
Display pg_stat_activity's wait_event when available

dlax:
## About pg_stat_activity's wait_event

* Before postgres 9.6, there's no `wait_event` nor `wait_event_type` column
* In postgres 9.6, `wait_event_type` values can be `LWLockNamed`, `LWLockTranche`, `Lock`, `BufferPin` and `wait_event` column is available.
* In postgres 10, new `wait_event_type` values: Activity, Extension, Client, IPC, Timeout, IO and many new `wait_event` (literal) values available.

Currently, in pg_activity:
*  prior to postgres 9.6, we simply use the `wait` (boolean) column of pg_stat_activity
* in postgres 9.6, we define the query as waiting if `wait_event IS NOT NULL`
* from version 10, we define the query as waiting when ` wait_event_type IN ('LWLock', 'Lock', 'BufferPin')`

(This seems inconsistent to me.)

## Proposed change

Always use `wait_event` when available and, if not NULL, declare the query as "waiting".

(See commit messages for extra details.)

### DIFF
--- a/pgactivity/activities.py
+++ b/pgactivity/activities.py
@@ -161,7 +161,7 @@ def sorted(processes: List[T], *, key: SortKey, reverse: bool = False) -> List[T
     ...         state="idle in transaction",
     ...         query="UPDATE pgbench_accounts SET abalance = abalance + 141 WHERE aid = 1932841;",
     ...         duration=0.1,
-    ...         wait=False,
+    ...         wait="ClientRead",
     ...         io_wait=False,
     ...         is_parallel_worker=False,
     ...     ),
--- a/pgactivity/queries/get_pg_activity_post_100000.sql
+++ b/pgactivity/queries/get_pg_activity_post_100000.sql
@@ -9,10 +9,7 @@ SELECT
           ELSE pg_stat_activity.client_addr::TEXT
       END AS client,
       EXTRACT(epoch FROM (NOW() - pg_stat_activity.{duration_column})) AS duration,
-      CASE WHEN pg_stat_activity.wait_event_type IN ('LWLock', 'Lock', 'BufferPin')
-          THEN true
-          ELSE false
-      END AS wait,
+      pg_stat_activity.wait_event as wait,
       pg_stat_activity.usename AS user,
       pg_stat_activity.state AS state,
       pg_stat_activity.query AS query,
--- a/pgactivity/queries/get_pg_activity_post_110000.sql
+++ b/pgactivity/queries/get_pg_activity_post_110000.sql
@@ -8,10 +8,7 @@ SELECT
           ELSE pg_stat_activity.client_addr::TEXT
       END AS client,
       EXTRACT(epoch FROM (NOW() - pg_stat_activity.{duration_column})) AS duration,
-      CASE WHEN pg_stat_activity.wait_event_type IN ('LWLock', 'Lock', 'BufferPin')
-          THEN true
-	  ELSE false
-      END AS wait,
+      pg_stat_activity.wait_event AS wait,
       pg_stat_activity.usename AS user,
       pg_stat_activity.state AS state,
       pg_stat_activity.query AS query,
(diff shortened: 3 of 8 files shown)


dlax:
@blogh what do you think of this idea?

dlax:
One assumption here is that we consider the query *waiting* whenever "pg_stat_activity.wait_event" is not NULL: is it valid?

blogh:
You're right to say it's not consistent.

First, I think it's important to be clear about what we want to report as "waiting" query.

In my opinion, we want to know queries that wait after other queries. (The one blocked by the blocking queries).

If we say that a query is waiting as soon as it has a wait event, we will have all the queries going in and out of the waiting queries all the time, because we're always waiting for something. And since most of these wait events being normal (eg : IO wait), the waiting query panel will be meaningless.

In my opinion  
* prior to postgres 9.6, we simply use the wait (boolean) column of pg_stat_activity => This is the correct behavior
* after 9.6 : report query as waiting as long as wait_event_type = 'Lock'

It would be nice to display the wait events.

dlax:
Not sure we should exclude all non-lock events. For instance, indicating 
that a process is waiting for the client (wait_event_type=Client) can 
also be useful, no? (In fact, that's the exact use-case I was trying 
address here "wait_event=ClientRead".)

blogh:
Why don't we just display the wait event an get rid of the wait column altogether ?

Could we switch the  "RUNNING QUERY" panel in "wait event mode" on a key press and display wait event info ?
It would be nice if we also had a column saying that the wait event has been there for the last X snapshot.

blogh:
Why don't we just display the wait event an get rid of the wait column altogether ?

Could we switch the  "RUNNING QUERY" panel in "wait event mode" on a key press and display wait event info ?
It would be nice if we also had a column saying that the wait event has been there for the last X snapshot.

dlax:
> Why don't we just display the wait event an get rid of the wait column altogether ?

That's exactly what this PR do. See how it renders for a pgbench session:
```
PID        TIME+          Waiting  IOW              state   Query                                       
5756    0.033036       ClientRead    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5752    0.031777      LockManager    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5759    0.028209    transactionid    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5761    0.014022                N    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5757    0.012355            tuple    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5749    0.011765            tuple    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5750    0.009352    transactionid    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5765    0.007795       ClientRead    N             active   UPDATE pgbench_branches SET bbalance = bbala
5754    0.006374    transactionid    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5766    0.005765    transactionid    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5764    0.004998    transactionid    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5760    0.003881            tuple    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5751    0.003325            tuple    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5755    0.002955    transactionid    N             active   UPDATE pgbench_branches SET bbalance = bbala
5758    0.002078            tuple    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5762    0.001422    transactionid    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5747    0.000799            tuple    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5748    0.000141       ClientRead    N      idle in trans   UPDATE pgbench_tellers SET tbalance = tbalan
5753    0.000000    transactionid    N             active   UPDATE pgbench_tellers SET tbalance = tbalan
5763    0.000000                N    N      idle in trans   BEGIN;
```

> Could we switch the "RUNNING QUERY" panel in "wait event mode" on a key press and display wait event info ?

Adding an extra view is possible, but this requires a fair amount of work, so I'd rather be sure it's worth. More specifically, I'd be interested in following the idea that not all wait events mean that the process is *abnormally* waiting and perhaps extract a "business" rule that we could document and use to display meaningful information.

blogh:
works great

dlax:
Yes, we still display Y/N when wait_event is not available (i.e old versions) because we have nothing available but a boolean status.

Issue #149, opened 2020-10-15
FATAL: 'utf-8' codec can't decode byte 0xe7 in position 128: invalid continuation byte


PatrickSiqueira:
Hello people!
I am currently using pg_activity-1.6.1 with PostgreSQL 12.3, after a certain run time, I get the following error:

Traceback (most recent call last):
  File "./pg_activity", line 332, in main
    procs, disp_procs)
  File "/var/lib/pgsql/install/pg_activity-1.6.1/pgactivity/UI.py", line 1191, in poll
    disp_proc)
  File "/var/lib/pgsql/install/pg_activity-1.6.1/pgactivity/UI.py", line 1337, in __poll_activities
    queries =  self.data.pg_get_activities(self.duration_mode)
  File "/var/lib/pgsql/install/pg_activity-1.6.1/pgactivity/Data.py", line 510, in pg_get_activities
    ret = cur.fetchall()
  File "/usr/lib64/python3.6/site-packages/psycopg2/extras.py", line 100, in fetchall
    res = super(DictCursorBase, self).fetchall()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe7 in position 128: invalid continuation byte
FATAL: 'utf-8' codec can't decode byte 0xe7 in position 128: invalid continuation byte

Previously with version 9 of PostgreSQL, I did not receive these errors, after the migration to version 12 is that I am having this. can you help me?

blogh:
Hi,

Are you using third party extensions on postgres ?
We had a similar issue related to powa before : https://github.com/dalibo/pg_activity/issues/130

Benoit

PatrickSiqueira:
Hello,

Only pg_stat_statements, but I've been using it since version 9 of postgres, could this problem be related to some incompatibility of pg_activity with postgres 12.3 or CentOS 8?

blogh:
Hi sorry for the long delay,

>  could this problem be related to some incompatibility of pg_activity with postgres 12.3 or CentOS 8?
I dont think so.

I'll try to reproduce this on my side.

Benoit.

dlax:
I could reproduce on a debian system with postgres 11 (from debian) by executing the following SQL while pg_activity is running:

```
$ psql postgres
denis@postgres=# CREATE DATABASE latin1 ENCODING 'latin1' TEMPLATE template0 LC_COLLATE 'fr_FR.latin1' LC_CTYPE 'fr_FR.latin1';
CREATE DATABASE
[...]

PR #212, opened 2021-05-10
Fix encoding errors #149 > still WIP

blogh:
When the encoding of a database is not UTF8. Queries with special
caracters might crash pg_activity with the message :

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 26:
++  invalid continuation byte

This patch fixes the issue by querying pg_database.encoding and using it
to encode the string.

### DIFF
--- a/pgactivity/queries/get_blocking.sql
+++ b/pgactivity/queries/get_blocking.sql
@@ -2,7 +2,7 @@
 SELECT
       pid,
       application_name,
-      datname AS database,
+      sq.datname AS database,
       usename AS user,
       client,
       relation,
@@ -17,17 +17,19 @@ SELECT
       END AS state,
       CASE WHEN sq.query LIKE '<IDLE>%%'
           THEN NULL
-          ELSE sq.query
+          ELSE convert_from(sq.query::bytea, coalesce(pg_catalog.pg_encoding_to_char(b.encoding), 'UTF8'))
       END AS query,
       waiting AS wait
   FROM
       (
+      -- Transaction id lock
       SELECT
             blocking.pid,
             '<unknown>' AS application_name,
             pg_stat_activity.current_query AS query,
             blocking.mode,
             pg_stat_activity.datname,
+            pg_stat_activity.datid,
             pg_stat_activity.usename,
             CASE WHEN pg_stat_activity.client_addr IS NULL
                 THEN 'local'
@@ -55,12 +57,14 @@ SELECT
                 ELSE extract(epoch from now() - {duration_column}) > %(min_duration)s
             END
       UNION ALL
+      -- VirtualXid Lock
       SELECT
             blocking.pid,
             '<unknown>' AS application_name,
             pg_stat_activity.current_query AS query,
             blocking.mode,
             pg_stat_activity.datname,
+            pg_stat_activity.datid,
             pg_stat_activity.usename,
             CASE WHEN pg_stat_activity.client_addr IS NULL
                 THEN 'local'
@@ -88,12 +92,14 @@ SELECT
                 ELSE extract(epoch from now() - {duration_column}) > %(min_duration)s
             END
       UNION ALL
+      -- Relation Lock
       SELECT
             blocking.pid,
             '<unknown>' AS application_name,
             pg_stat_activity.current_query AS query,
             blocking.mode,
             pg_stat_activity.datname,
+            pg_stat_activity.datid,
             pg_stat_activity.usename,
             CASE WHEN pg_stat_activity.client_addr IS NULL
                 THEN 'local'
@@ -125,18 +131,20 @@ SELECT
                 ELSE extract(epoch from now() - {duration_column}) > %(min_duration)s
             END
       ) AS sq
+      LEFT OUTER JOIN pg_database b ON sq.datid = b.oid
 GROUP BY
       pid,
       application_name,
-      query,
+      database,
+      usename,
+      client,
+      relation,
       mode,
       locktype,
       duration,
-      datname,
-      usename,
-      client,
       state,
-      relation,
-      wait
+      query,
+      encoding,
+      wait_event
 ORDER BY
       duration DESC;
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 import asyncio
 import pathlib
 import threading
+from typing import Optional
 
 import psycopg2
 import psycopg2.errors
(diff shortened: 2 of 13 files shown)


blogh:
There is probably a clever way to do this ... will install an Ubuntu 20 next week.

dlax:
@blogh, terminating the blocking query cleans things up for following tests, see added commit. This is a workaround, but I did not find a better solution.

Can you rebase the branch on master and possibly squash fixup commits? I would have a couple of additional comments then.

blogh:
@dlax: thx for the fix. rebase & sqash done.

pgactivity/queries/get_blocking.sql
```diff
@@ -125,18 +131,19 @@ SELECT
...
+--      query,
+--      mode,
+--      locktype,
+--      duration,
+--      database,
+--      usename,
+--      client,
+--      state,
+--      relation,
+--      wait
```
dlax, line 147: Should we keep these commented lines?

pgactivity/queries/get_blocking.sql
```diff
@@ -17,17 +17,19 @@ SELECT
...
   FROM
       (
+      -- Transaction id lock
       SELECT
             blocking.pid,
             '<unknown>' AS application_name,
             pg_stat_activity.current_query AS query,
             blocking.mode,
             pg_stat_activity.datname,
+	    pg_stat_activity.datid,
```
dlax, line 32: There appears to be a leading tab here (and everywhere else below); can you use spaces for consistency?

tests/conftest.py
```diff
@@ -21,9 +21,14 @@ def execute(postgresql):
 
     loop = asyncio.new_event_loop()
 
-    def execute(query: str, commit: bool = False, autocommit: bool = False) -> None:
+    def execute(
+        query: str, commit: bool = False, autocommit: bool = False, dbname: str = None
```
dlax, line 25: Should be `Optional[str] = None` (`Optional` to be imported from `typing`.)

tests/test_data.py
```diff
@@ -112,6 +113,49 @@ def test_terminate_backend(postgresql, data):
...
+    execute("select pg_sleep(2)")
     time.sleep(1)
     assert data.pg_get_active_connections() == 2
+
+
+def test_encoding(postgresql, data, execute):
+    """Test for issue #149"""
+    postgresql.set_session(autocommit=True)
+    with postgresql.cursor() as cur:
+        """plateform specific locales (,Centos, Ubuntu)"""
```
dlax, line 138: Should be a comment (`# ...`).

tests/test_data.py
```diff
@@ -112,6 +113,49 @@ def test_terminate_backend(postgresql, data):
...
+    execute(
+        "INSERT INTO tbl(s) VALUES ('initilialized éléphant')",
+        dbname="latin1",
+        commit=True,
+    )
+    execute("UPDATE tbl SET s = 'blocking éléphant'", dbname="latin1")
+    execute("UPDATE tbl SET s = 'waiting éléphant'", dbname="latin1", commit=True)
+    time.sleep(2)
+    running = data.pg_get_activities()
+    assert "éléphant" in running[0].query
```
dlax, line 161: I'd suggest `assert "blocking éléphant" in running[0].query` and same below because all queries actually contain `éléphant` so it'd be easier to distinguish them.

tests/test_data.py
```diff
@@ -6,6 +6,22 @@
 from pgactivity.data import Data
 
 
+def wait_for_data(
+    fct, timeout: int = 2, desc_recursionError: str = "timeout in wait_for_data"
```
dlax, line 10: I'd suggest `msg` instead of `desc_recursionError`; and drop the default value since all callers pass a value apparently.

tests/test_data.py
```diff
@@ -6,6 +6,22 @@
...
+    count = int(timeout / 0.1)
+    for _ in range(count):
+        time.sleep(0.1)
+        data = fct()
+
+        if not data:
+            continue
+        break
+    else:
+        raise RecursionError(desc_recursionError)
```
dlax, line 21: `AssertionError` would be more appropriate.

dlax (changes requested):
LGTM overall; a couple of style changes suggested though.

pgactivity/queries/get_blocking.sql
blogh: According to the [pg_locks docs](https://www.postgresql.org/docs/14/view-pg-locks.html)
> A single process can be waiting to acquire at most one lock at a time.

Therefore: 

* the blocking => blocked join can return only one row (at most). 
* blocked => pg_stat_activity will return one row only

The `group by` seems useless to me.

I will remove the block as suggested.

blogh: According to the [pg_locks docs](https://www.postgresql.org/docs/14/view-pg-locks.html)
> A single process can be waiting to acquire at most one lock at a time.

Therefore: 

* the blocking => blocked join can return several rows. 
* blocked => pg_stat_activity will return one row only

It's a mistake. I will put it back and amend the test.

tests/test_data.py
blogh: Fixed in 6e080de820bfff697476e45d0bcb69898cc967a6

tests/test_data.py
blogh: Fixed in 6e080de820bfff697476e45d0bcb69898cc967a6

blogh:
will fix everything in different commits and squach after if you re fine with the changes.

tests/test_data.py
blogh: done in 1381d25ebd9792762cbf4903aab1e856ee752a1a

tests/test_data.py
```diff
@@ -112,6 +115,51 @@ def test_terminate_backend(postgresql, data):
...
+        commit=True,
+    )
+    execute("UPDATE tbl SET s = 'blocking éléphant'", dbname="latin1")
+    execute("UPDATE tbl SET s = 'waiting éléphant'", dbname="latin1", commit=True)
+    running = wait_for_data(data.pg_get_activities, msg="could not fetch activities")
+    assert "blocking éléphant" in running[0].query
+    (waiting,) = data.pg_get_waiting()
+    assert "éléphant" in waiting.query
+    (blocking,) = data.pg_get_blocking()
+    assert "éléphant" in blocking.query
```
dlax, line 155: `waiting éléphant` and `blocking éléphant` here as well :)

blogh: right https://github.com/dalibo/pg_activity/pull/212/commits/4ec554aa74e463ed04d0b9c3f95347d3036b1b0a

blogh: thx

tests/conftest.py
blogh: done in edc56aa

tests/test_data.py
blogh: done in 0584de4

pgactivity/queries/get_blocking.sql
blogh: Fixed this on + some in get_blocking_post_90200.sql and get_pg_activity.sql in  d1140fc

pgactivity/queries/get_blocking.sql
blogh: Fixed the `group by` thing  in f5818a4

dlax (approved):
LGTM; I'd suggest to squash (all or some) commits to get a cleaner history. Thanks!

blogh:
Thx for the review / help

Issue #208, opened 2021-05-03
random spacing issue in output


plockaby:
I'm running pg_activity 2.1.0 from the postgres debian apt repository. It's putting spaces at random on query strings and make it very hard to read what is going on. This only started happening with 2.1.0.

<img width="413" alt="Screen Shot 2021-05-03 at 10 22 04 AM" src="https://user-images.githubusercontent.com/633173/116909852-a4802500-abf9-11eb-8e33-509ebd3b3297.png">

dlax:
> This only started happening with 2.1.0.

Which version were you using before?

I'd need more information about your environment: operating system, terminal emulator, font configuration in the terminal, locale and encoding information.

plockaby:
Ah, very good questions. Sorry for the lack of information. Here's what I have so far.

The previous version that seemed to work OK was 1.4.0-1 from the Debian 10 "buster" archive. The version that I'm running now, that seems to have a problem, is on Debian 10 "buster", all against PostgreSQL 13. I'm using version `2.1.0-1.pgdg100+1` installed from `deb http://apt.postgresql.org/pub/repos/apt/ buster-pgdg main`. I tried a few different configurations and this is what I've come up with. I use iTerm2 as my primary terminal emulator. The font is Monaco Regular 11, anti-aliased. I've also seen it happen on macOS Terminal on both Big Sur and Catalina.

```
TERM=xterm-256color
LANG=en_US.UTF-8
```

I was also seeing it happen in this mode:

```
TERM=xterm-256color
LANG=C.UTF-8
```

Further testing was kind of all over the place. On one server I got it to happen very consistently at a terminal width of 144 but not 143. On another server 144 was a-ok but terminal width 117 caused the display issues.

I manually installed pg_activity 2.0.0 from pip into a virtual environment and it experiences the same issue. When I installed pg_activity 1.4.0 in a virtual environment the problem did not appear.

I hope that provides some more context. Let me know if there is anything else that I can provide.

dlax:
If I'm reading correctly, you're using a terminal on macOS but pg_activity is running on a Debian system, so I suppose you're using an SSH connection, right?

It'd then be helpful to try either directly on the Debian server or use an SSH connection from another (preferably Linux) system.

For the record, pg_activity was almost completely rewritten in version 2.0, so quite a few things have changed (noticeably, we're now using a third-party library, [blessed](https://blessed.readthedocs.io/), for terminal rendering).

plockaby:
[...]

PR #214, opened 2021-05-12
Do not use blessed's text wrapper for the query string

dlax:
This apparently poses problem for some terminals
(https://github.com/dalibo/pg_activity/issues/208) and we, in fact, do
not need a sequence-aware text wrapper here, since the query string
contains no escape sequence, but only text.

### DIFF
--- a/pgactivity/views.py
+++ b/pgactivity/views.py
@@ -1,6 +1,6 @@
 import functools
 import inspect
-from textwrap import dedent
+from textwrap import TextWrapper, dedent
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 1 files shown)


pgactivity/views.py
```diff
@@ -74,7 +74,7 @@ def shorten(term: Terminal, text: str, width: Optional[int] = None) -> str:
     """
     if not text:
         return ""
-    wrapped: List[str] = term.wrap(text, width=width, max_lines=1)
+    wrapped = term.wrap(text, width=width, max_lines=1)
```
dlax, line 77: Note to myself: left-over (but probably needed) `term.wrap()` here.

PR #217, opened 2021-05-12
Add tests for the "data" module

dlax:
Should make it easier to add regression tests for #149

### DIFF
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -1,0 +1,92 @@
+import time
+
+import pytest
+
+from pgactivity.data import Data
+
+
+@pytest.fixture
+def data(postgresql):
+    return Data.pg_connect(
+        1,
+        host=postgresql.info.host,
+        port=postgresql.info.port,
+        database=postgresql.info.dbname,
+        user=postgresql.info.user,
+    )
+
+
+def test_pg_is_local(postgresql, data):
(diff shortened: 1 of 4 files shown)


tests/test_data.py
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+
+def test_activities(postgresql, data, execute):
+    with postgresql.cursor() as cur:
+        cur.execute("SELECT pg_sleep(1)")
+    (running,) = data.pg_get_activities()
+    assert "pg_sleep" in running.query
+    assert running.state == "idle in transaction"
+    assert running.wait == "ClientRead"
```
blogh, line 56: We need a version test here since the output is different before pg 10: 

```suggestion
    if postgresql.server_version >= 100000:
        assert running.wait == "ClientRead"
    else:                                                                       
        assert running.wait == None
```

dlax: This should not have slipped in since we're supposed to run tests with postgres version 9.6 to 13. But there's a configuration mistake, so this wasn't true. Adding a fix in the PR, thanks.

dlax: Hm, in fact, it's totally broken and fixing this is beyond the scope of the PR so I'll disable the invalid CI setup here until I find time to improve it.

PR #218, opened 2021-05-13
Fix bug in wrapping long queries

plockaby:
This addresses #208


dlax:
Merged by hand as https://github.com/dalibo/pg_activity/commit/e44cab8ba14b3bb664ae1133f8c915e1027fcb01; many thanks for working on this!

PR #219, opened 2021-05-14
Add a waitevent column to blocking queries

blogh:
Blocking queries can be long for a number of reasons, the wait_event
column of pg_activity can help us understand them. Therefore, it was
added to the blocking query panel.

### DIFF
--- a/docs/man/pg_activity.pod
+++ b/docs/man/pg_activity.pod
@@ -73,7 +73,7 @@ B<min duration> seconds. It displays the following information:
 
 =item - B<TIME>: time since the beginning of the query / transaction / backend start depending on the B<DURATION_MODE> currently in use;
 
-=item - B<Waiting>: specific wait event if available. Otherwise, a boolean indicating that we are waiting for a Lock;
+=item - B<Waiting>: for PostgreSQL 9.6+: a specific wait event or nothing. Otherwise, a boolean indicating if we are waiting for a Lock;
 
 =item - B<IOW>: boolean indicating that the process is waiting for IO as reported by the psutil library;
 
@@ -138,6 +138,8 @@ requiered by another session. It shows following information:
 =item - B<MODE>: the mode of the lock;
 
 =item - B<TIME+>: the duration of the query, transaction or session depending on the B<DURATION_MODE> setting;
+
+=item - B<Waiting>: for PostgreSQL 9.6+: a specific wait event or nothing. Otherwise, a boolean indicating if we are waiting for a Lock;
 
 =item - B<state>: the state of the transaction;
 
--- a/pgactivity/data.py
+++ b/pgactivity/data.py
@@ -16,7 +16,7 @@ from psycopg2.errors import (
 from psycopg2.extensions import connection
 
 from . import queries
-from .types import BWProcess, RunningProcess
+from .types import BlockingProcess, WaitingProcess, RunningProcess
 from .utils import clean_str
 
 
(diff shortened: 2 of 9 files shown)


blogh:
This PR is still lacking the man page update.
I'll wait for #216  to be merged before adding it.

blogh:
Tests would also have to be added when #217 is merged.

blogh:
Man page updated

pgactivity/types.py
```diff
@@ -702,7 +703,7 @@ class RunningProcess(BaseProcess):
 
 
 @attr.s(auto_attribs=True, frozen=True, slots=True)
-class BWProcess(BaseProcess):
+class WaitingProcess(BaseProcess):
     """Process for a blocking or waiting query."""
```
dlax, line 707:
```suggestion
    """Process for a waiting query."""
```

pgactivity/queries/get_blocking.sql
```diff
@@ -18,7 +18,8 @@ SELECT
       CASE WHEN sq.query LIKE '<IDLE>%%'
           THEN NULL
           ELSE sq.query
-      END AS query
+      END AS query,
+      NULL AS wait
```
dlax, line 22: Why not using `pg_stat_activity.waiting`?

docs/man/pg_activity.pod
```diff
@@ -139,6 +139,8 @@ requiered by another session. It shows following information:
 
 =item - B<TIME+>: the duration of the query, transaction or session depending on the B<DURATION_MODE> setting;
 
+=item - B<Waiting>: specific wait event if available. Otherwise, a boolean indicating it we are waiting for a Lock;
```
dlax, line 142:
```suggestion
=item - B<Waiting>: specific wait event if available. Otherwise, a boolean indicating if we are waiting for a Lock;
```

(Besides, it's not quite accurate that "we are waiting for a Lock" and it's not always a boolean either. It's unrelated to this PR, but we might want to update this afterwards.)

dlax (changes requested):

pgactivity/queries/get_blocking.sql
blogh: no reason .. will fix that way

blogh:
Fixed the mistakes and tried to reword the description of the waiting column in the man page

pgactivity/data.py
```diff
@@ -278,13 +278,15 @@ def pg_get_waiting(self, duration_mode: int = 1) -> List[BWProcess]:
...
+    def pg_get_blocking(self, duration_mode: int = 1) -> List[BlockingProcess]:
         """
         Get blocking queries
         """
-        if self.pg_num_version >= 90200:
+        if self.pg_num_version >= 100000:
+            query = queries.get("get_blocking_post_100000")
+        elif self.pg_num_version >= 90200:
             query = queries.get("get_blocking_post_90200")
         elif self.pg_num_version < 90200:
```
dlax, line 291: For clarity, I think we should turn this `elif` into an `else`.

dlax (changes requested):
One minor nit, LGTM otherwise.

pgactivity/data.py
blogh: You mean the last one ? if so should I fix the other instances of this in data.py ?

dlax: Yes. If there are more of these, I'd suggest to clean this up in a dedicated commit as this is unrelated to this change.

blogh: done

blogh:
* Fixed the if .. elif ..else
* Fixed a mystake on my part wait_event where introduced in 9.6 not 10 (But only for lock / lwlocks / bufferpin)
  * modified the manpage
  * modified version checks for pg9.6 and query file
  
 Test pass from 9.6 to 13.

dlax (approved):

Issue #213, opened 2021-05-11
Revamp the header


blogh:
Currently, we have those infos in the header part : (the 3 last lines are removed in non local mode)

```
PostgreSQL 13.2 - benoit-dalibo - postgres@/var/run/postgresql:5433/postgres - Ref.: 2s
 Size:       845.73M - 0B/s           | TPS:                     1         | Active connections:          1 | Duration mode: query
 Mem.:      35.6% - 2.12G/7.46G       | IO Max:                  0/s
 Swap:      4.9% - 575.23M/11.39G     | Read:           0B/s - 0/s
 Load:         0.06 0.10 0.09         | Write:          0B/s - 0/s
```

Detail :
- Size : Size of the database you are connected to  - Growth in B/s
- Mem. : Pct used - memory used / memory available (in local mode only)
- Swap : Pct swap used - swap used / swap available (in local mode only)
- Load : CPU load  (in local mode only)
- TPS  : Transaction per second (sum of commit & rollback for all databases / time elapsed since last snap)
- IO Max : Displays the iops (in local mode only)
- Read : Current Read thruput  (aggregated data from the psutil library) (in local mode only)
- Write : Current Write thruput (aggregated fata from the psutil library) (in local mode only)
- Active connections : count of active connections
- Duration mode : query/transaction/backend (seems out of place)

It would be nice to have more:

- Active connections : count of active connections / max_connections (superuser_reserved_connections)
- Idle :  count of idle connection / max_connections (superuser_reserved_connections)
- Idle in transaction :  count of idle in transaction connections / max_connections (superuser_reserved_connections)
- Idle in transaction aborted :  count of idle in transaction aborted connections / max_connections (superuser_reserved_connections)
- worker process : count of worker processes /  max_worker_processes
- parallel workers : count of parallel workers used for query processing  / max_parallel_workers
- maintenance workers : count of parallel workers used for maintenance / max_parallel_maintenance_workers
- logical workers : count of  workers used for logical replication / max_logical_replication_workers
- autovacuum workers : count of autovacuum workers / autovacuum_max_workers

We also could list the number of wal senders / wal receiver. But I am not convinced that it's usefull in the header. I am also not fond of having logical workers but they are taken from the worker process pool (like parallel workers and maintenance workers) so not having them here could make the info incomplete / misleading.

Note: the availability of these infos depend on the version of PostgreSQL

Example :

```
[...]

PR #220, opened 2021-05-18
Overhaul the header view

dlax:
We rework the header view in order to make the rendering more compact,
by limiting the amount of white spaces between columns, and to make the
columns / rows data more extensible.

Now the header() view displays two blocks:

* first row (containing "Size", "Active connections", "Duration mode"
  information) is always visible and has its own alignment
* subsequent rows, only available in local mode and displaying extra
  system information, also use their own alignment

In the first block, we change the | column delimiter into a blue ⋅. In
the second one, we use a comma (similarly to what top uses).

Besides, columns no longer have a fixed width, thus make the rendering
more compact.

Adding extra columns should be simpler now, thus preparing work for #213.

### DIFF
--- a/pgactivity/views.py
+++ b/pgactivity/views.py
@@ -1,5 +1,6 @@
 import functools
 import inspect
+import itertools
 from textwrap import TextWrapper, dedent
 from typing import (
     Any,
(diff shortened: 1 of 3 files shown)


blogh:
LGTM

PR #224, opened 2021-05-24
Blocking query now lists transactions which blocks virtualxid.

sahapasci:
I had a long running transaction which blocked a repack process. I realized that pg_activity couldn't display which ended up with this patch. 

I could only test get_blocking_post_90600.sql since I don't have any old environment.

### DIFF
--- a/pgactivity/queries/get_blocking.sql
+++ b/pgactivity/queries/get_blocking.sql
@@ -75,6 +75,39 @@ SELECT
             pg_locks AS blocking
             JOIN (
                 SELECT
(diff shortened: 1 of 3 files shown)


blogh:
@sahapasci good catch. 

would you mind if I "push --force" the tests in your master branch ? (some stuff as been added to our master branch in the mean time so it needs a rebase)

blogh:
Merged will do another PR to add the tests

sahapasci:
@blogh  sure no problem

PR #244, opened 2021-09-27
use threads for queries launched in //

blogh:
* this enables us to do the following without locking ourselves
   * run blocking
   * run blocked
   * run blocked
* the wait_for_data is useless now because we wait until we are
  we are locked or idle_in_transaction before resuming the execution
* there is a cleanup at the end of the tests to avoid edge effects
* adds a test with multiple blockers
* changes the default for min_duration 1 => 0
* updates test_data and test_ui to use the new API

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,15 @@
-import asyncio
 import pathlib
-import threading
-from typing import Optional
+import time
+from threading import Thread
+from typing import Dict, List, Optional
 
 import psycopg2
 import psycopg2.errors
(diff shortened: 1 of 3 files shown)


blogh:
This is what we talked about.
It's probably over overengineered  but I would be glad to have tips if there's something obviously bloken / bad.

blogh:
Looks likeit works on my computer but not online...

blogh:
ah no ok I get the same at home.
I guess my thread are never finished when I get an assertion.

Issue #239, opened 2021-09-14
the latest release is very slow


bagafoot:
I'm using pg_activity 2.2.1 release, when scroll up/down its response quite slow compared to previous releases.
It is not big deal but a little bit annoying.

dlax:
Which previous release(s) do you compare to?

bagafoot:
1.5

dpiscitelli:
Hello,

Sadly, I have to confirm this point, slower than 1.6 in my environments

blogh:
With the PR #247, scrolling should work properly.
I think I found why it's slow and will address it in a separate PR.

blogh:
The performance should be better with #249.
There is still a bug related to scrolling with wrapping / ident + wrapping.

dlax:
I think we can close this issue as performance problems have been resolved. The scrolling issues will be tracked in #260.

PR #247, opened 2021-10-14
Fix scrolling issues

blogh:
The PR aims to solve some problems related to scrolling (#239)


blogh:
Will add / fix tests later.

PR #249, opened 2021-10-20
Improve performance during refresh

blogh:
Scrolling is very slow and a long press on up/down is not very manageable.


blogh:
Done with `functools.lru_cache()` in the PR.

PR #252, opened 2021-11-09
Fix IOW status

gozdal:
The `disk sleep` had been used which seems to have a typo (` ` instead of `-`).
Safer to use a constant from psutil.


dlax (approved):
CI failures are due to a dependency upgrade.

gozdal:
That was quick! 👍

PR #254, opened 2021-12-02
Document authentication

miohtama:
How to pass the password to the PSQL connection is not obvious from the current documentation. 
Add a basic example with links to the more documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,3 +153,16 @@ reporting are still accounted for by postgres in `pg_stat_get_db_xact_commit()`
 and `pg_stat_get_db_xact_commit()`. Therefore `pg_activity` will display a non
 zero TPS even with no activity on the database, and/or no activity displayed on
 screen.
+
+**How can I specify a password for authentication ?**
+
+pg_activity uses libpq to access to PostgreSQL therefore all the traditional
+methods are available.
+
+You can pass the password for the database connection in a password file.
+Information can also be given via PostgreSQL's environment variables
+(PGPASSFILE or PGPASSWORD) or via the connection string parameters.
+
+The password file is preferred since it's more secure (security is deferred to
+the OS). Please avoid password in connection strings at all cost.
+


README.md
```diff
@@ -125,6 +125,14 @@ Navigation mode
 | `q`        | Quit                                          |
 | `Other`    | Back to activity                              |
 
+Authentication
+--------------
+
+You can pass the password for the database connection in `PGPASSWORD` 
+in environment variable, or other 
+[PostgreSQL environment variable](https://www.postgresql.org/docs/current/libpq-envars.html)
+supported by `psycopg2` library.
```
dlax, line 134: It's not specific to psycopg2, it's just libpq:
```suggestion
supported by `libpq`.
```

dlax (changes requested):
@miohtama, thank you for your contribution!

@blogh, what do you think about this addition?

blogh:
This is a libpq feature and there are many different ways to do this.

I'd rather have this in the FAQ part and would prefer something like :

> pg_activity uses `libpq` to access to PostgreSQL. All the traditional methods are available. 
>
> You can pass the password for the database connection in a [password file](https://www.postgresql.org/docs/current/libpq-pgpass.html). Information can also be given
> via [PostgreSQL's environment variables](https://www.postgresql.org/docs/current/libpq-envars.html) (PGPASSFILE or PGPASSWORD) or via the [connection string parameters](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS).
>
> The password file is preferred since it's more secure (security is deferred to the OS). Please avoid password in connection string at all cost.

blogh:
Also : @miohtama, thank you for your contribution :)

blogh:
Are you guys ok with this ? If so I can do the PR.

dlax:
> Are you guys ok with this ? If so I can do the PR.

Yes.

blogh:
Pushed it that way. 

@miohtama, thanks for the patch
@dlax, thanks for reviewing

PR #257, opened 2021-12-13
Fix scrolling when wrapping is involved

blogh:
With a wrapped display mode the current display is bugged. If some
queries are wrapped before the line we put the focus on. It's possible
that the said line is out of the screen.

With this patch, the line with the focus is the always the first one.

When a the focus is on a selected line, the text is now bold with a
lighter brand of cyan. It is otherwise difficult to see if the first
ligne of the display is selected since it always has the focus.


dlax:
There is something wrong when scrolling with arrows: the list of processes decreases until it gets restored when getting empty.

tests/test_ui.txt
```diff
@@ -996,7 +995,6 @@ PostgreSQL ... - test - postgres@127.0.0.1:.../tests - Ref.: 2s
  Size: 106.07M - 0B/s ⋅ TPS: 0 ⋅ Active connections: 2 ⋅ Duration mode: query
                                 RUNNING QUERIES
 PID    DATABASE                      state   Query
-...    tests                 idle in trans   SELECT 43
```
dlax, line 999: Why these changes in tests? Is it related to the scrolling issue mentioned in my other comment?

blogh: yes

blogh:
The active line is always on the top of the screen.
In a way, pressing "down" moves all the line up. (it's ugly but I don't see how to do it otherwise with the current @limit machinery.)
When all line are exhausted, we go back to the first line. (we could remove this.)

blogh:
Replaced by #268

PR #258, opened 2021-12-13
Revamp the infos in the header

blogh:
This patch aims to expand and revamp the information present in the
header. The informations is now split in tree groups.

Instance info:
* Add more instance informations: utptime, and revamp the existing info;
* Add session status info;
* can be toggle with the i key.

Worker info:
* Add the worker count info: for parallelism / logical replication but also
  autovacuum, walsender, and walreceiver;
* can be toggle with the w key.

System info:
* Reformat the sytem information.
* can be toggle with the s key.

Any combination of theses groups can be active at a given time.

(related to see #213)


blogh:
Still WIP, with no test yet.

dlax (commented):
Great work so far. Just noted a few things after a quick read but no real blocker :)

I'll do a more in-depth review when this will be ready on your side.

blogh:
I wonder if it's necessary to have two separate functions + queries for `pg_get_db_info` and `pg_get_server_activity`, while we could easily do everything in one go (since everything is used only in the header)  and fetch everything into a single object.

pgactivity/activities.py
```diff
@@ -228,13 +229,29 @@ def get_mem_swap() -> Tuple[float, int, int, float, int, int]:
     buffers = getattr(phymem, "buffers", 0)
     cached = getattr(phymem, "cached", 0)
     mem_used = phymem.total - (phymem.free + buffers + cached)
-    return (phymem.percent, mem_used, phymem.total, vmem.percent, vmem.used, vmem.total)
+    ram_pct_bc = round(100 * (buffers + cached) / phymem.total, 1)
+    ram_pct_free = round(100 * phymem.free / phymem.total, 1)
+    swap_pct_free = round(100 * vmem.free / vmem.total, 1)
```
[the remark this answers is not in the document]
blogh: Do you mean creating a percentage type and perform the rounding in the render function ?

pgactivity/data.py
```diff
@@ -217,23 +217,31 @@ def pg_get_db_info(
...
             "size_ev": size_ev,
+            "uptime": ret["uptime"],
+            "cache_hit_ratio": ret["cache_hit_ratio"],
         }
 
-    def pg_get_active_connections(self) -> int:
+    def pg_get_server_activity(self) -> ServerActivityDict:
         """
-        Get total of active connections.
+        Get server acticity (session, workers, cache hit ratio etc..)
```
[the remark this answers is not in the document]
blogh: added to the next commit.

pgactivity/activities.py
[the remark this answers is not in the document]
dlax: Yes, exactly. That'd be consistent with other `.percent` values already returned.

dlax:
> I wonder if it's necessary to have two separate functions + queries for `pg_get_db_info` and `pg_get_server_activity`, while we could easily do everything in one go (since everything is used only in the header) and fetch everything into a single object.

Agreed.

pgactivity/views.py
```diff
@@ -199,55 +229,105 @@ def header(
...
         )
     )
 
-    total_size = utils.naturalsize(dbinfo.total_size)
-    size_ev = utils.naturalsize(dbinfo.size_ev)
-
-    def render_columns(columns: List[List[str]], *, delimiter: str) -> Iterator[str]:
-        column_widths = [
-            max(len(column_row) for column_row in column) for column in columns
+    total_size = utils.naturalsize(si.total_size)
```
blogh, line 238: Should all thoses natural size calls be moved to the server_informations class ? (generate a total_size_natural or add a method that does this in the class ? or an methode disguised as an attribute)

Or should I avoid this and keep a distinction between the data and it's visualization ?

pgactivity/views.py
```diff
@@ -184,12 +169,57 @@ def header(
...
+
+    @render.register(str)
+    def render_str(s: str) -> str:
+        return term.bold_green(s)
+
+    @render.register(int)
+    def render_int(n: int) -> str:
+        return term.bold_green(str(n))
+
+    @render.register(Pct)
```
blogh, line 191: This doesn't work. My `Pct` objects are seen as `floats`.
Is the approach flawed ? Is there a way to "copy" the class instead of inheriting it ?

pgactivity/types.py
```diff
@@ -607,20 +611,42 @@ class Host:
...
+    pct_free: Pct = attr.ib(init=False)
 
     @classmethod
     def default(cls) -> "MemoryInfo":
-        return cls(0.0, 0, 0)
+        return cls(0, 0, 0, 0)
+
+    def __attrs_post_init__(self) -> None:
+        # prevents Frozen=True
+        self.pct_used = self.used * 100 / self.total
```
blogh, line 647: I moved the percent calculus here instead of in the `mem_swap_load()` function (in activities.py).

Is it the proper way to d it ?

pgactivity/views.py
dlax: Yes, that's fine as it is now; keep this logic at rendering time.

pgactivity/types.py
dlax: Since these percentage values are computed, they should be made available as properties of the class (where the computation will always be perform). This was, we're certain that the percentage value is always consistent with underlying data.

pgactivity/views.py
dlax: Seems to work on my side with the following example code:
```python
from functools import singledispatch                                
from typing import Any               
                                     
class pct(float):                    
    pass                             
                                     
@singledispatch                      
def fun(arg: Any) -> str:            
    return "any"                     
                                     
@fun.register(pct)                   
def fun_pct(arg: pct) -> str:        
    return "pct"                     
                                     
@fun.register(float)                 
def fun_float(arg: float) -> str:    
    return "float"                   
                                     
@fun.register(int)                   
def fun_int(arg: int) -> str:        
    return "int"                     
                                     
assert fun.dispatch(float) is fun_float
assert fun.dispatch(pct) is fun_pct
assert fun.dispatch(int) is fun_int

assert fun(pct(1.2)) == 'pct'
assert fun(0.1) == 'float'
```

But I see below that `render_pct()` and `render_float()` have the same implementation...

blogh: One has a "%" the other has not.

pgactivity/types.py
blogh: Ok I'll rewrite it with __setattr__ to make the class frozen again.

pgactivity/views.py
blogh: didn't pay attention to the assert with the functions thats neat ...

pgactivity/views.py
blogh: Sorry for asking again, I want to make sure I understand the rationale.
Is it better to :

- keep the logic in different places ?
- keep the execution of the logic during rendering time ? in that case, would a property do the job ?

pgactivity/views.py
dlax: > One has a "%" the other has not.

Indeed ;)

So does it work?

pgactivity/views.py
dlax: It is better to keep data untransformed in the `ServerInformation` object and then format it here in the view. I don't think a property is worth because it'll be only used once.

The logic is:
* get data from the database
* possibly transform it into something that makes sense for our application (rounding does not count as such, but statistics would *if* we don't store the underlying data -- for instance, you don't store the mean value of a series along with the series itself, otherwise the statistics may get inconsistent with actual data)
* store raw data in a live object
* render data from that object in a view, where "render" can mean rounding, simple calculation, etc.

pgactivity/types.py
dlax: No, not `__setattr__`, this is equivalent to setting an attribute.
Rather:
```python
@property
def pct_used(self) -> float:
     return self.used * 100 / self.total
```

blogh: Ok I understand now with the explanation above. thx.

pgactivity/views.py
blogh: thx for the explanation.

pgactivity/views.py
blogh: Nope. Will look into why it's not working asap.

blogh: well, that's lovely. 
I changed it from my `object.__setattr__` to the `property` and it works.

tests/test_data.py
```diff
@@ -41,16 +41,8 @@ def test_pg_is_local_access(postgresql, data):
...
-    dbinfo = data.pg_get_db_info(None)
-    assert set(dbinfo) == {
-        "timestamp",
-        "no_xact",
-        "total_size",
-        "max_length",
-        "tps",
-        "size_ev",
-    }
+def pg_get_server_informations(data):
```
blogh, line 44: Don't really know how to go about this given : 

1. the amount of data to test (not a real excuse, I am just being lazy)
2. the variance between (supported) versions
3. the mess it's gona be to create meaningful tests for each and every info (when it's even possible) AND keep them functional across PostgreSQL version (some are optimize specific).

pgactivity/types.py
```diff
@@ -698,25 +729,118 @@ def default(
...
+        # prevents Frozen=True
+        if self.parallel_workers is None and self.logical_workers is None:
+            self.worker_processes = None
+        else:
+            self.worker_processes = (
+                0 if self.parallel_workers is None else self.parallel_workers
+            ) + (0 if self.logical_workers is None else self.logical_workers)
+
+    @classmethod
+    def default(cls) -> "ServerInformations":
```
dlax, line 805: Is this method actually used?

tests/test_data.py
dlax: Maybe we can keep the same simple sanity check as before (i.e. only make sure that we have expected fields)?

blogh: I think that's what I did.

We used to return all the fields to the caller and do the object creation there. 
I thought it was cleaner to return the object directly. So now if we're missing filed in the select
the object creation will fail.

Some fields might be `None` because across versions some info are missing.

pgactivity/types.py
blogh: Nope, I think I didn't get how the default methods where used so I added it.
And by the time, I understand how they where used, I thought it had always been there...

I'll remove it thx !

tests/test_data.py
dlax: Sorry, I think I misunderstood your question given the diff context. Can you reformulate?

pgactivity/types.py
blogh: done.

pgactivity/types.py
```diff
@@ -685,28 +716,79 @@ def default(
...
+    tuples_returned: int
+    insert_per_second: int
+    update_per_second: int
+    delete_per_second: int
+    tuples_returned_per_second: int
+    temp_files: int
+    temp_bytes: int
+    max_dbname_length: int
+    waiting: int
+    worker_processes: int = attr.ib(init=False)
```
blogh, line 780: I guess this is not very safe since the class is not Frozen ? or is it ?

tests/test_data.py
blogh: dealt with irl

pgactivity/types.py
dlax: Yes, this should be a property. Better if the class can be made frozen too.

blogh:
Rebased.

pgactivity/ui.py
```diff
@@ -229,11 +237,9 @@ def main(
                     term,
                     ui,
                     host=host,
-                    dbinfo=dbinfo,
                     pg_version=data.pg_version,
-                    tps=tps,
-                    active_connections=active_connections,
-                    activity_stats=activity_stats,
+                    server_information=server_information,
+                    activity_stats=activity_stats,  # FIXME
```
dlax, line 242: What's the `FIXME` for?

dlax (changes requested):
Suggesting minor improvements. LGTM overall.

Commits should be squashed before merging, also.

pgactivity/types.py
```diff
@@ -16,6 +16,7 @@
     TypeVar,
     Union,
 )
+from datetime import timedelta
```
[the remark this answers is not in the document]
blogh: the import at the beginning of activities should be changed also ?
```
1   import builtins
  1 import os
  2 import time
  3 from typing import Dict, List, Optional, Sequence, Tuple, TypeVar
  4 from warnings import catch_warnings, simplefilter
  5 
  6 import attr << move up ?
  7 import psutil << move up ?
  8 
  9 from .types import (
 10     BlockingProcess,
 11     WaitingProcess,
 12     IOCounter,
 13     LoadAverage,
 14     LocalRunningProcess,
 15     MemoryInfo,
 16     RunningProcess,
 17     SortKey,
 18     SystemProcess,
 19     SwapInfo,
 20 )
```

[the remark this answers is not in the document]
blogh: In retrospect, I also dont undestand these from `types.py`, should I move them up ?

```
   2 import attr
   1 import psutil
23   from attr import validators
```

pgactivity/types.py
```diff
@@ -24,9 +25,13 @@
 from . import colors, utils
 
 
-T = TypeVar("T")
+class Pct(float):
+    """Used to distinguish percentage from float when displaying the header"""
+
+    pass
```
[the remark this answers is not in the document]
blogh: Done, thx.

pgactivity/ui.py
blogh: I wasn't super fond of the tuple we use there.
But I don't have anything better to propose.

Removed !

pgactivity/types.py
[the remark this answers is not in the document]
dlax: The import order is:
1. builtins or `__future__`
2. imports from standard library
3. imports from dependencies
4. imports from our modules

And within each kind of imports:
1. `import <module>`
2. `from <module> import <name>`

So here, `attr` and `psutil` are not part of standard library, so they should be located after stdlib imports.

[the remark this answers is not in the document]
dlax: That being said, if things are inconsistent already, these should be not fixed as part of this PR but separately. For this, we might use [isort](https://pycqa.github.io/isort/).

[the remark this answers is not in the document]
blogh: Nice tool, thanks for the explanation.

pgactivity/ui.py
```diff
@@ -16,7 +16,12 @@ def main(
     options: optparse.Values,
     dsn: str,
     *,
-    render_header: bool = True,
+    render_header: types.HeaderVisibilityInfo = types.HeaderVisibilityInfo(
```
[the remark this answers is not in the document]
blogh: I added some parameters to hide the header information.

If it's ok with you I will add the info to the man page.

pgactivity/types.py
```diff
@@ -685,28 +762,82 @@ def default(
...
+    @property
+    def worker_processes(self) -> int:
+        if self.parallel_workers is None and self.logical_replication_workers is None:
+            return None
+        else:
+            return (0 if self.parallel_workers is None else self.parallel_workers) + (
+                0
+                if self.logical_replication_workers is None
+                else self.logical_replication_workers
+            )
```
dlax, line 838: There's something wrong here or in fields definition above because most (all?) fields involved here are not declared as `Optional[int]` type so conditions like `self.parallel_workers is None` will never be `True`.

Maybe fields type should be update to, e.g., `Optional[int]`?

Then the property apparently returns `Optional[int]` as well.

tests/test_data.py
```diff
@@ -41,16 +41,8 @@ def test_pg_is_local_access(postgresql, data):
...
-    dbinfo = data.pg_get_db_info(None)
-    assert set(dbinfo) == {
-        "timestamp",
-        "no_xact",
-        "total_size",
-        "max_length",
-        "tps",
-        "size_ev",
-    }
+def pg_get_server_information(data):
```
dlax, line 44:
```suggestion
def test_pg_get_server_information(data):
```

(Only functions named `test_<something>` are collected as tests.)

dlax (changes requested):

pgactivity/types.py
blogh: The typing is wrong and it's true for a good bunch of them.

blogh: I changed : 
* several attributes to Optional[int].
* the output type for the method.

Added a query file for pre 9.0 versions when I realised max_wal_senders was optional because back then there was no replication. Theses version are untested (I only have builds from 9.0 onwards) but the breakage was obvious so I fixed it.

blogh: Is it better ?

dlax: Yes, better; thanks!

dlax (commented):
I've just added an extra commit to hide the "i" key from help when not in local mode.

LGTM otherwise. Can you rework the git history before we merge this?

blogh:
Added the new parameters in the man page and fixed the "process" name there (=> "worker process").
Squashed.

blogh:
I think it's good to merge now. 

I dont get which review change It's complaining about...

dlax (approved):

PR #268, opened 2022-01-11
Keep focused line on top of processes table when query is wrapped

dlax:
This is an alternative proposal to #257 which (in my opinion) produces a less strange behavior.

When scrolling through processes and the query is wrapped, there is no
way to compute an offset for the table rows in order to move the cursor.
Instead, when query is wrapped, we now always display focused process on
top of the table and we wrap around the processes list (display first
processes at the end of the table).

@blogh, what do you think?


blogh:
Cool, I like it.

blogh (approved):

PR #269, opened 2022-01-11
Do not display 'None' values in processes table

dlax:
Instead do not display anything (i.e. use an empty string when
rendering).


blogh (approved):
Nice

PR #270, opened 2022-01-17
Rename query file, add comment and fixes

blogh:
* Renames all query files so that they can be sorted.
* Add comments about what is the reason for the creation of this query
  file (what is new in this version).
* Fix some missing filters in the "oldest query" files.


blogh:
Tested for 9.0 to 14 (pytest for 9.6+, quick test for errors for older versions)

dlax (approved):

PR #271, opened 2022-01-20
README.md: howto execute development versions for non-devs

Krysztophe:
1) Installation instructions of the dev version for people with low python skills, like me (thanks @blogh).

2) I've added the name of the packages from the PGDG. The one from Debian is not so obvious.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ This is only necessary to test development versions. First, clone the repository
 
     $ git clone https://github.com/dalibo/pg_activity.git
 
-Change the branch if necessary. Then create an dedicated environment,
+Change the branch if necessary. Then create a dedicated environment,
 install dependencies and then pg\_activity from the repo:
 
     $ python3 -m venv MYENV
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ along with psycopg2 (version 2.8 or higher):
 
     $ python3 -m pip install pg_activity psycopg2-binary
 
-Then launch:
+In case your `$PATH` does not already contain it, the full path is:
 
     $ ~/.local/bin/pg_activity
 
--- a/README.md
+++ b/README.md
@@ -44,16 +44,16 @@ This is only necessary to test development versions. First, clone the repository
 Change the branch if necessary. Then create a dedicated environment,
 install dependencies and then pg\_activity from the repo:
 
-    $ python3 -m venv MYENV
-    $ source MYENV/bin/activate
-    $ python3 -m pip install psycopg2-binary
-    $ python3 -m pip install pg_activity/
-    $ pg_activity
+    $ cd pg_activity
+    $ python3 -m venv .venv
+    $ . .venv/bin/activate
+    (.venv) $ pip install psycopg2-binary .
+    (.venv) $ pg_activity
 
 To quit this env and destroy it:
 
     $ deactivate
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,20 +8,53 @@ Command line tool for PostgreSQL server activity monitoring.
...
+    $ ~/.local/bin/pg_activity
+
+Installation from the git repository
+------------------------------------
+
+This is only necessary to test development versions. First, clone the repository:
+
+    $ git clone https://github.com/dalibo/pg_activity.git
+
+Change the branch if necessary. Then create an dedicated environment,
```
blogh, line 44:
```suggestion
Change the branch if necessary. Then create a dedicated environment,
```

README.md
```diff
@@ -8,20 +8,53 @@ Command line tool for PostgreSQL server activity monitoring.
...
+install dependencies and then pg\_activity from the repo:
```
blogh, line 45:
```suggestion
install pg_activity and it's dependencies. Finally, run pg\_activity from the 
virtual environment:
```

README.md
```diff
@@ -8,20 +8,53 @@ Command line tool for PostgreSQL server activity monitoring.
...
+---------------------
 
 Alternatively, pg\_activity can be installed using pip on Python 3.6 or later
 along with psycopg2 (version 2.8 or higher):
 
     $ python3 -m pip install pg_activity psycopg2-binary
 
+Then launch:
+
+    $ ~/.local/bin/pg_activity
```
blogh, line 35: Not sure the info is needed since calling pg_activity works just fine as is.

blogh: Thanks for  the PR !

README.md
Krysztophe: Dependencies first, isn'it ?

And « its », not « it's »

README.md
Krysztophe: When installing through pip  with the postgres user on Ubuntu, I have this:

```
  WARNING: The script pg_activity is installed in '/var/lib/postgresql/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
```

I suppose that `.local/bin` is not in the `$PATH` of system users.

README.md
```diff
@@ -8,20 +8,53 @@ Command line tool for PostgreSQL server activity monitoring.
...
+    $ git clone https://github.com/dalibo/pg_activity.git
+
+Change the branch if necessary. Then create a dedicated environment,
+install dependencies and then pg\_activity from the repo:
+
+    $ python3 -m venv MYENV
+    $ source MYENV/bin/activate
+    $ python3 -m pip install psycopg2-binary
+    $ python3 -m pip install pg_activity/
+    $ pg_activity
```
dlax, line 51: I'd recommended the following:
```console
$ cd pg_activity
$ python3 -m venv .venv
$ . .venv/bin/activate
(.venv) $ pip install . psycopg2-binary
(.venv) $ pg_activity
```

- `.venv` is an usual name for virtualenvs
- it's quite common to put the virtualenv inside the source tree
- when the venv is activated, one can use `pip` directly
- and install from current directory

Krysztophe: Thanks!  PR updated

dlax (approved):

PR #273, opened 2022-01-20
Change cache hit ratio to reflect change since last snap

### LINKED ISSUE #272: Hit ratio is not an instantaneous metric
In the new headers, I have a hit ratio stuck à 99%, changing very slowly.  I suppose that it comes from pg_database.

I would expect it to be an average on a few seconds, like other metrics in the header.

blogh:
This fixes #272


blogh:
@Krysztophe : can you try this ?

Krysztophe:
As said in private:

with a pgbench scale 1000 (about the size of the 12 Gb share buffers), after a server restart, I get immediately a hit ratio around 65%-90%, although only a very small part of the tables are in the SB (checked with pg_buffercache) ?!?

blogh:
I tried with scale 300 (4GB) and the default shared buffers.
I got similar results.
But I also see the same with `hit_ratio` service of `check_pgactivity`.

pgactivity/data.py
```diff
@@ -232,11 +233,13 @@ def pg_get_server_information(
             )
             ret = cur.fetchone()
 
+        hr: Optional[Pct] = Pct(0.0)
```
dlax, line 236: If you initialize to 0 here, can the value actually be `None`?

pgactivity/views.py
```diff
@@ -242,7 +242,7 @@ def indent(text: str) -> str:
         columns = [
             [f"* Global: {render(uptime)} uptime"],
             [f"{render(total_size)} dbs size - {render(size_ev)} growth"],
-            [f"{render(si.cache_hit_ratio)} cache hit ratio"],
+            [f"{render(si.cache_hit_ratio_last_snap)} cache hit ratio"],
```
dlax, line 245: How does this render if `None`? (Not sure that can happen, per comment above.)

blogh: It will render as "-" because the renderer will match NoneType.

https://github.com/dalibo/pg_activity/blob/b42264982b93d9052372338ca24ac3fec6b4145f/pgactivity/views.py#L174-L177

dlax: Ah, I didn't remember that; that's great!

pgactivity/data.py
blogh: When I compute the value if it didn't change since last time it's set to None.

blogh: It felt weird to set the hit ratio to 100% or 0% in that case.

blogh: maybe it's worth a comment ?

dlax: Rather, I'd suggest to rewrite the `hr` assignment below without the ternary `if else`; it'd be clearer.

blogh: Done

docs/man/pg_activity.pod
```diff
@@ -38,7 +38,7 @@ Global :
 
 =item - B<growth>: groth in B/s of the databases (filtered);
 
-=item - B<cache hit ratio>: the percentage of page read from the PostgreSQL's cache (filtered).
+=item - B<cache hit ratio>: the percentage of page read from the PostgreSQL's cache since last snapshot(filtered).
```
dlax, line 41:
```suggestion
=item - B<cache hit ratio>: the percentage of page read from the PostgreSQL's cache since last snapshot (filtered).
```

dlax (changes requested):

blogh:
I have regened the man file and squashed the commits.

dlax (approved):
LGTM

PR #276, opened 2022-01-21
Move from optparse to argparse

blogh:


dlax:
This is python, not java; so `optparse`, not `optParse` (likewise for `argparse`) :)

pgactivity/cli.py
```diff
@@ -2,104 +2,103 @@
...
-def get_parser() -> OptionParser:
-    parser = OptionParser(
-        add_help_option=False,
-        version="%prog " + __version__,
-        usage="%prog [options] [connection string]",
+def get_parser() -> ArgumentParser:
+    parser = ArgumentParser(
+        usage="%(prog)s [options] [connection string]",
+        description=(
+            "htop like application for PostgreSQL " "server activity monitoring."
```
dlax, line 17:
```suggestion
            "htop like application for PostgreSQL server activity monitoring."
```

pgactivity/cli.py
```diff
@@ -2,104 +2,103 @@
...
-        description=(
-            "htop like application for PostgreSQL " "server activity monitoring."
-        ),
+        add_help=False,
+    )
+    # Connection string
+    parser.add_argument(
+        "connection_string",
+        help="A valid connection string to the database, e.g.: "
+        "'host=HOSTNAME port=PORT user=USER dbname=DBNAME'.",
```
dlax, line 32:
```suggestion
        help=(
            "A valid connection string to the database, e.g.: "
            "'host=HOSTNAME port=PORT user=USER dbname=DBNAME'."
        ),
```

tests/test_ui.txt
```diff
@@ -89,7 +89,7 @@ Default CLI options, passed to ui.main():
 ...         return ks
 ...
 ...     with patch.object(term, "inkey", new=inkey):
-...         ui.main(term, dataobj, host, options, "",
+...         ui.main(term, dataobj, host, options,
```
dlax, line 92: It seems we're still using optparse in this test (see above, out of diff context).

In fact, I believe we should get rid of argparse/optparse when entering application code from the `cli` module. This can be done by transforming `argparse.Namespace` into a `dict` calling `ui.main()` with `vars(args)`. This change is probably more invasive, so I'd suggest to do it in a dedicated commit.

dlax (changes requested):
Nice step forwards, thanks!

As I noticed that we have an option group to control columns in the table, I am wondering if it'd make sense to also define a group for connection options. (Might be done later, of course.)

So apart from `optparse`'s left-over in tests, this looks good to me.

pgactivity/cli.py
```diff
@@ -2,104 +2,103 @@
...
             "only supported from libpq 9.2)"
         ),
-        description=(
-            "htop like application for PostgreSQL " "server activity monitoring."
-        ),
+        add_help=False,
+    )
+    # Connection string
+    parser.add_argument(
+        "connection_string",
```
dlax, line 30: I'd suggest to add `metavar="connection string"` so that the help of this argument will be consistent with the usage line.

blogh: done

tests/test_ui.txt
blogh: yes ... completely missed this part ... should have waited monday to push ...

blogh: I changed it but I am not sure I did it right.

blogh:
I added a group for the connection stuff and changed the tests.
I'll squash the patch and fix the spelling thing in the git message, once It's ok for you.

pgactivity/cli.py
```diff
@@ -137,127 +97,187 @@ def get_parser() -> OptionParser:
...
+        "--dbname",
+        dest="dbname",
+        help="Database name to connect to",
+        metavar="DBNAME",
+    )
 
-    group = OptionGroup(
-        parser, "Display Options, you can exclude some columns by using them "
+    group = parser.add_argument_group(
+        "Display Options", "you can exclude some columns by using them."
```
dlax, line 163:
```suggestion
        "Display Options", "These options may be used hide some columns from the processes table."
```

tests/test_ui.txt
dlax: Improved this in the commit I just pushed. (Basically use an `argparse.Namespace` object directly. The overall diff is smaller.)

dlax (approved):
One minor rephrase suggestion. Ok otherwise.

tests/test_ui.txt
blogh: yes ... indeed. I had the feeling I was writing way to much code for this ...

blogh: thanks.

blogh:
erf sorry forgot to clean the history....

PR #291, opened 2022-05-09
Fix parallel workers under the lead process in the GUI

kmoppel-cognite:
Currently they're jumping around annoyingly. Relevant for vers >= 9.6
where parallel query was added

### DIFF
--- a/pgactivity/activities.py
+++ b/pgactivity/activities.py
@@ -157,6 +157,23 @@ def sorted(processes: List[T], *, key: SortKey, reverse: bool = False) -> List[T
     ...         user="postgres",
     ...         client="local",
     ...         cpu=0.1,
+    ...         mem=0.994_254_939_413_836,
+    ...         read=0.1,
+    ...         write=0.282_725_318_098_656_75,
+    ...         state="idle in transaction",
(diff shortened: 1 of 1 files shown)


blogh:
I think the sort has to be done in two steps.
We sort in reverse order for the time and the other metrics but we have to sort in ascending order for the pid.

cf https://docs.python.org/3/howto/sorting.html#sort-stability-and-complex-sorts

kmoppel-cognite:
@blogh I think you're right...need some custom sort function...but then...why not to do it in SQL still as I proposed in the first patch - currently actually I think there are no alternative view modes for "Running queries" where rows can be re-ordered in the UI ?

blogh:
... of course I forgot mypy ... after black ... 
I really should code more often than once every 5 months ...
fixing it ...

dlax (approved):
Fixed/changed a few minor things based on @blogh commit, rebased and squashed.

LGTM as is; the parallel worker issue can be addressed in a separate changeset.

dlax:
> I think the way parallel processes are handled is just broken

Opened https://github.com/dalibo/pg_activity/issues/297 to track this.

PR #295, opened 2022-06-01
Ignore all __pycache__ folders

kianmeng:
Running `tox -p` command produced these supposed to be ignored folders:

pgactivity/queries/__pycache__/
tests/__pycache__/

We also sort the ignored list in .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,8 @@
-build/
-htmlcov/
-dist/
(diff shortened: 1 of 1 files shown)


kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #296, opened 2022-06-08
Add a --refresh option to the cli

### LINKED ISSUE #293: change refresh rate via argument on startup
Please, could you add argument for delay (refresh rate) like other monitoring tools have (e.g. `htop` has `-d`) - I've read man pages and there is only + or - once already being in UI, but no argument on invoking the script...

blogh:
Adds a `--refresh` parameter to the cli as suggested in the issue #293
`-d` is already used for the database name.

Closes #293.


pgactivity/cli.py
```diff
@@ -97,6 +97,16 @@ def get_parser() -> ArgumentParser:
         metavar="FIELD:REGEX",
         default=[],
     )
+    # --refresh
+    parser.add_argument(
+        "--refresh",
+        dest="refresh",
+        help=("Refresh frequency."),
```
dlax, line 104: Superfluous parentheses.

dlax (approved):

pgactivity/cli.py
dlax: Also, in man page it's called "Refresh rate", not "frequency"; perhaps make the naming uniform (slightly preferring "rate" over "frequency")?

blogh: I applied your proposals and amended the previous patch.

Then I went into a uniformisation spree in a separate commit. (should probably be a different PR ?)

blogh:
rebased + back mypy pytest run

dlax (approved):
Just fixed a case issue in help message and improved `--refresh` help text in an extra commit.

LGTM

PR #305, opened 2022-09-02
Add logging to pgactivity

blogh:
* adds a stream logger gets info level messages and prints them when
  we exit.
* adds a file logger gets all the messages and logs them to a file
  specified by the new `--debug-file` option if available.

### DIFF
--- a/pgactivity/cli.py
+++ b/pgactivity/cli.py
@@ -1,13 +1,46 @@
+import logging
 import os
 import socket
 import sys
 import time
 from argparse import ArgumentParser
+from io import StringIO
+from typing import Optional
 
 from blessed import Terminal
 from psycopg2.errors import OperationalError
 
 from . import __version__, data, types, ui
+
+
+def configure_logger(debug_file: Optional[str] = None) -> StringIO:
+    logger = logging.getLogger("pgactivity")
+    logger.setLevel(logging.DEBUG)
+
+    # The steamhandler is used to print hints to the user at exit.
+    # The INFO log level is reserved for this.
+    memory_string = StringIO()
+    c_handler = logging.StreamHandler(memory_string)
+    c_handler.setLevel(logging.INFO)
+    c_handler.name = "stream_handler"
+
+    c_format = logging.Formatter("Hint - %(message)s")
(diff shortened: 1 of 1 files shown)


pgactivity/cli.py
```diff
@@ -1,15 +1,48 @@
...
 from psycopg2.errors import OperationalError
 
 from . import __version__, data, types, ui
 
 
+def configure_logger(debug_file: Optional[str] = None) -> StringIO:
+    logger = logging.getLogger("pgactivity")
+    logger.setLevel(logging.DEBUG)
+
+    # The steamhandler is used to print hints to the user when he exists.
```
dlax, line 20:
```suggestion
    # The steamhandler is used to print hints to the user at exit.
```

pgactivity/cli.py
```diff
@@ -356,10 +398,10 @@ def main() -> None:
...
                     dataobj = newdataobj
-                    print(term.clear + term.home, end="")
                     break
         except KeyboardInterrupt:
             sys.exit(1)
         else:
-            print(term.clear + term.home, end="")
             break
+        finally:
+            print(memory_stream.getvalue())
```
dlax, line 407: Printing to stdout is intended? or should we use stderr?

dlax (approved):

pgactivity/cli.py
blogh: It was an oversight. Fixed.

pgactivity/cli.py
```diff
@@ -1,15 +1,48 @@
...
 
 from . import __version__, data, types, ui
 
 
+def configure_logger(debug_file: Optional[str] = None) -> StringIO:
+    logger = logging.getLogger("pgactivity")
+    logger.setLevel(logging.DEBUG)
+
+    # The steamhandler is used to print hints to the user at exit.
+    # The INFO log level is reserved for this.
```
dlax, line 21: "Reserving" a log level for some purpose is rather fragile (in general, less here since we control everything). A better design would be define a custom logger for this:
```python
hint_logger = logging.getLogger("pgactivity-hints")
```

PR #307, opened 2022-09-05
Add hint messages when features are disabled.

blogh:
This patch adds hint messages to give feedback to the user when a
feature is disabled. Namely :

* tempfile (disabled because of privileges or timeout)
* wal receiver (disabled on Aurora because the feature not available)
* system counters (disabled because of lack of privilege, feature unavailable
  on Mac, access to the pid file)

### DIFF
--- a/pgactivity/data.py
+++ b/pgactivity/data.py
@@ -1,4 +1,5 @@
 import getpass
+import logging
 import re
 from argparse import Namespace
 from typing import Dict, List, Optional
@@ -32,6 +33,9 @@ from .types import (
     NO_FILTER,
 )
 from .utils import clean_str
+
+
+logger = logging.getLogger("pgactivity")
 
 
 def pg_get_version(pg_conn: connection) -> str:
@@ -145,27 +149,41 @@ class Data:
         Verify if the user running pg_activity can access
         system information for the postmaster process.
         """
-        try:
-            query = queries.get("get_pid_file")
-            with self.pg_conn.cursor() as cur:
-                # This query doesn't crash when the user doesn't have the
-                # required privilege to acces the data_directory guc
-                # it will just return an empty string
+        query = queries.get("get_data_directory")
+        with self.pg_conn.cursor() as cur:
+            try:
                 cur.execute(query)
-                ret = cur.fetchone()
-            pid_file = ret["pid_file"]
+            except InsufficientPrivilege:
+                logger.info(
+                    "Insufficient privilege to show data_directory. System counters are disabled."
+                )
+                return False
+            ret = cur.fetchone()
+
+        pid_file = f"{ret['data_directory']}/postmaster.pid"
+        try:
             with open(pid_file, "r") as fd:
-                pid = fd.readlines()[0].strip()
-                try:
-                    proc = psutil.Process(int(pid))
-                    proc.io_counters()
-                    proc.cpu_times()
-                    return True
-                except psutil.AccessDenied:
-                    return False
-                except Exception:
-                    return False
-        except Exception:
+                pid = fd.readline().strip()
+        except PermissionError:
+            logger.info("Access Denied to the pidfile. System counters are disabled.")
+            return False
+
+        try:
+            proc = psutil.Process(int(pid))
+            proc.io_counters()
+            proc.cpu_times()
+            return True
+        except psutil.AccessDenied:
+            logger.info(
+                "Access denied to the psutil data. System counters are disabled."
+            )
+            return False
+        except AttributeError:
+            # See issue #300
+            logger.info(
+                "Your platform doesn't support some of psutil features required to get system counters. "
+                "System counters are disabled."
+            )
             return False
 
     def pg_cancel_backend(self, pid: int) -> bool:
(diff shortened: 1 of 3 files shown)


blogh:
I changed `pg_is_local_access()`. The exceptions are more targeted. I think it's better ... but if I missed one it will crash.

Do you think I should keep the `except Exception` we had ?

pgactivity/data.py
```diff
@@ -216,13 +229,20 @@ def pg_get_temporary_file(self) -> Optional[TempFileInfo]:
                 # superuser or pg_read_server_files are required (Issue #278)
                 cur.execute(queries.get("reset_statement_timeout"))
                 self.failed_queries.temp_file_query_failed = True
+                logging.getLogger("pgactivity").info(
+                    "Insufficient privilege to fetch the tempfile data. The feature was disabled. Please use --no-tempfile or a platform specific setting (eg. --rds)."
```
dlax, line 233:
```suggestion
                    "Insufficient privilege to fetch the tempfile data. The feature was disabled. Please use --no-tempfiles or a platform specific setting (eg. --rds)."
```

pgactivity/data.py
```diff
@@ -315,15 +338,23 @@ def pg_get_server_information(
...
+                    query,
+                    {
+                        "dbname_filter": self.filters.dbname,
+                        "skip_db_size": skip_db_size,
+                        "prev_total_size": prev_total_size,
+                        "using_rds": using_rds,
+                    },
+                )
+            except InsufficientPrivilege as e:
+                if e.pgerror.find("permission denied for database") >= 0:
```
dlax, line 352: Would this work if the error message is localized? Or is it not?

pgactivity/queries/get_data_directory.sql
```diff
@@ -0,0 +1,2 @@
+-- Get the path of the pidfile
+SHOW data_directory;
```
dlax, line 2: `SHOW` would not work when we'll move to using psycopg 3. Perhaps we can keep the query to `pg_settings`?

dlax (changes requested):

pgactivity/data.py
```diff
@@ -145,28 +146,40 @@ def pg_is_local_access(self) -> bool:
...
+                )
+                return False
+            ret = cur.fetchone()
+        try:
+            pid_file = f"{ret['data_directory']}/postmaster.pid"
+        except PermissionError:
+            logger.info("Access Denied to the pidfile. System counters are disabled.")
             return False
+        with open(pid_file, "r") as fd:
+            pid = fd.readlines()[0].strip()
```
dlax, line 166: You can change this as:
```suggestion
            pid = fd.readline().strip()
```
by the way.

pgactivity/data.py
```diff
@@ -145,28 +146,40 @@ def pg_is_local_access(self) -> bool:
...
+            try:
```
dlax, line 167: At this point the file can be closed, so you can dedent the following statements.

pgactivity/queries/get_data_directory.sql
blogh: I like this because we have an error.
The query returns nothing when we don't have the privilege.

I can revert to the previous query.

dlax: Sorry, I'm wrong actually. I mixed up `SHOW` and `SET`. `SHOW` should work.

pgactivity/data.py
```diff
@@ -145,27 +149,41 @@ def pg_is_local_access(self) -> bool:
...
+            return False
+
+        try:
+            proc = psutil.Process(int(pid))
+            proc.io_counters()
+            proc.cpu_times()
+            return True
+        except psutil.AccessDenied:
+            logger.info(
+                "Acces denied to the psutil data. System counters are disabled."
```
dlax, line 178:
```suggestion
                "Access denied to the psutil data. System counters are disabled."
```

pgactivity/data.py
```diff
@@ -145,27 +149,41 @@ def pg_is_local_access(self) -> bool:
...
+            )
+            return False
+        except AttributeError:
+            # See issue #300
+            logger.info(
+                "You platform doesn't support some of psutil's features required to get system counters. "
```
dlax, line 184:
```suggestion
                "Your platform doesn't support some of psutil features required to get system counters. "
```

pgactivity/data.py
```diff
@@ -264,6 +291,10 @@ def pg_get_wal_receivers(self) -> Optional[int]:
         except FeatureNotSupported:
             # Not implemented on Aurora (Issue #301)
             self.failed_queries.wal_receivers_query_failed = True
+            logger.info(
+                "The receiver information is not available on you platform. "
```
dlax, line 295:
```suggestion
                "The receiver information is not available on your platform. "
```

dlax:
> Do you think I should keep the `except Exception` we had ?

No, that's fine. It doesn't seem needed any more.

dlax (approved):
Just a few typos; LGTM otherwise.

PR #309, opened 2022-09-06
Fix the parallel worker position in duration sort mode

### LINKED ISSUE #297: processes sort with parallel worker is broken
See discussion at #291.

From postgres 13, maybe we could rely on `leader_pid` column of `pg_stat_activity` view as first sort step on SQL side?

blogh:
We now sort by `query_leader_pid` ASC, `is_parallel_worker` ASC in
addition to `duration` DESC.

`query_leader_pid` is a crafted column:
* pre pg13, it's always `NULL`, the tuple order will still depend on the
  order in witch the tuple are fetched;
* for pg13 and after, it's equal to `coaelesce(leader_pid, pid)` which
  should fix the display.

Fix some tests which displayed different queries for a leader and it's
worker processes.

closes #297


dlax (approved):
Excellent!

PR #315, opened 2022-09-19
Update README.md

blogh:
The main page has not been updated for a while. This patch updates it with new:

* parameters
* hotkeys
* description
* screenshot


README.md
```diff
@@ -61,58 +61,71 @@ Usage
 `pg_activity` works locally or remotely. In local execution context, to obtain
 sufficient rights to display system information, the system user running
 `pg_activity` must be the same user running postgresql server (`postgres` by
-default), or have more rights like `root`. Otherwise, `pg_activity` can fallback
-to a degraded mode without displaying system information. On the same way,
-PostgreSQL user used to connect to the database must be super-user.
+default), or have more rights like `root`. The PostgreSQL user used to connect
+to the database must be super-user in order to get as much data as possible.
+Otherwise, `pg_activity` can fallback to a degraded mode where some data like
```
dlax, line 66:
```suggestion
Otherwise, `pg_activity` can fall back to a degraded mode where some data like
```

dlax (changes requested):
The new screenshot seems very wide and I fear it'd be hard to read when included in the README which renders quite narrow. Perhaps you can resize your terminal before doing the screenshot so that the process rows fill the whole height? Also, can you exclude the terminal menu bar?

blogh:
I retook the screen with another color palette and reduce the overall size.

dlax:
The image is blurry; and there is still the menu and scroll bar from your terminal.

Sorry to be picky on this, but I took some care to make the previous screenshot look nice and I'd thus appreciate if we keep the same "quality"...

dlax:
@blogh, I have pushed here a screenshot I just made; it's similar to the previous one, using `--no-app-name --no-user --no-client` options so that it's narrower. It's indeed wider than the previous one, but I don't think we can work around this as there is no much more information to display. I have verified that it renders fine on github within the README.

dlax (approved):

dlax:
@blogh, if you're okay with it, please go ahead and merge (possibly squashing commits).

blogh:
thanks for the fix

PR #317, opened 2022-09-21
Colorize the "duration mode"

dlax:
It used to be in green, then the color got dropped during header refactoring. Now using yellow, to make it less visible than more important information.


blogh:
Do you want to highlight it because it's a user input ?
If so, we could also highlight the refresh rate.

dlax:
Ok; also colorized "Ref." and "Min. duration" fields.

blogh (approved):
Nice, thanks.

PR #324, opened 2022-10-14
Fix pidfile not found error

blogh:


pgactivity/data.py
```diff
@@ -165,7 +165,12 @@ def pg_is_local_access(self) -> bool:
             with open(pid_file, "r") as fd:
                 pid = fd.readline().strip()
         except PermissionError:
-            logger.info("Access Denied to the pidfile. System counters are disabled.")
+            logger.info(
+                f"Access Denied to the pidfile: {pid_file}. System counters are disabled."
+            )
+            return False
+        except FileNotFoundError:
+            logger.info(f"pidfile not found: {pid_file}. System counters are disabled.")
```
dlax, line 173: We could catch `OSError` and log the specific error:
```python
except OSError as e:
     logger.info("pidfile %s could not be read: %s. System counters are disabled.", pid_file, e)
    return False
```

Also note that we should not format the log message ourselves as it possibly does superfluous work if the logging level is higher than `INFO` here.

dlax (changes requested):

pgactivity/data.py
blogh: > Also note that we should not format the log message ourselves as it possibly does superfluous work if the logging level is higher than INFO here.

... clever ... :)

dlax (approved):

scoopex:
Yes please :-)


Git this today from ubuntu :-(
(connection established by a kubernetes forward)
```
export PGPASSWORD=yoloesee
export PGUSER=postgres
export PGDATABASE=postgres
export PGHOST=localhost
pg_activity
  File "/usr/bin/pg_activity", line 33, in <module>
    sys.exit(load_entry_point('pg-activity==3.0.1', 'console_scripts', 'pg_activity')())
  File "/usr/lib/python3/dist-packages/pgactivity/cli.py", line 407, in main
    ui.main(term, dataobj, host, args)
  File "/usr/lib/python3/dist-packages/pgactivity/ui.py", line 27, in main
    is_local = data.pg_is_local() and data.pg_is_local_access()
  File "/usr/lib/python3/dist-packages/pgactivity/data.py", line 165, in pg_is_local_access
    with open(pid_file, "r") as fd:
FileNotFoundError: [Errno 2] No such file or directory: '/bitnami/postgresql/data/postmaster.pid'

``

blogh:
Thanks for reporting.
It seems that this PR doesn't solve everything (see ##323). Does it fix the problem in your case ?

dlax:
> FileNotFoundError: [Errno 2] No such file or directory: '/bitnami/postgresql/data/postmaster.pid'

I think this PR would resolve this issue.

Issue #275, opened 2022-01-21
Encoding bug, still not out of the woods ...


blogh:
We had some encoding issues there : #149. 
Well we're not out of the woods yet ... there are still some edge cases (at least one) .

In psql :
```
pcsql -c "SELECT '\It`s gona crash', pg_sleep(10)"
```

Quickly :
```
(pgactivity) [benoit@benoit-dalibo pg_activity]$ pg_activity 
Traceback (most recent call last):
  File "/home/benoit/tmp/pgactivity/bin/pg_activity", line 33, in <module>
    sys.exit(load_entry_point('pg-activity', 'console_scripts', 'pg_activity')())
  File "/home/benoit/git/dalibo/pg_activity/pgactivity/cli.py", line 312, in main
    ui.main(term, dataobj, host, options, dsn)
  File "/home/benoit/git/dalibo/pg_activity/pgactivity/ui.py", line 200, in main
    pg_procs.set_items(data.pg_get_activities(ui.duration_mode))
  File "/home/benoit/git/dalibo/pg_activity/pgactivity/data.py", line 278, in pg_get_activities
    cur.execute(
  File "/home/benoit/tmp/pgactivity/lib64/python3.9/site-packages/psycopg2/extras.py", line 146, in execute
    return super(DictCursor, self).execute(query, vars)
psycopg2.errors.InvalidTextRepresentation: invalid input syntax for type bytea
```

It's because of this part of the queries pg_get_activity*.sql: 
```
convert_from(a.query::bytea, coalesce(pg_catalog.pg_encoding_to_char(b.encoding), 'UTF8')) AS query,
```

In order for the cast to work, the '\' should be escaped:
```
[local]:5435 postgres@postgres=# SELECT convert_from('\It`s gona crash', 'UTF8');
ERROR:  invalid input syntax for type bytea
LINE 1: SELECT convert_from('\It`s gona crash', 'UTF8');
                            ^
[local]:5435 postgres@postgres=# SELECT convert_from(E'\It`s gona crash', 'UTF8');
  convert_from   
-----------------
 It`s gona crash
(1 row)
[local]:5435 postgres@postgres=# SELECT convert_from('\\It`s gona crash', 'UTF8');
   convert_from   
------------------
 \It`s gona crash
(1 row)
```

ssharunas:
This bug crashes pg_activity if there are a lot of queries with regexp or newlines.
Example:
```SQL
select '123' ~ '\d+', pg_sleep(3);
[...]

PR #331, opened 2023-01-10
Add test for #275

blogh:
This test will ensure we still to handle queries without throwing this exception :

psycopg2.errors.InvalidTextRepresentation: invalid input syntax for type bytea

### DIFF
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -137,6 +137,13 @@ def test_encoding(postgresql, data, execute):
     assert "blocking éléphant" in blocking.query
 
 
+def test_InvalidTextRepresentation(postgresql, data, execute):
+    """Test for issue #275"""
+    postgresql.execute("select '123' ~ '\\d+', 'Hello world!\n', pg_sleep(3)")
+    running = data.pg_get_activities()
+    assert "123" in running[0].query
+
+
 def test_filters_dbname(data, execute):
     data_filtered = attr.evolve(data, filters=types.Filters(dbname="temp"))
     execute("SELECT pg_sleep(2)", dbname="template1", autocommit=True)


tests/test_data.py
```diff
@@ -137,6 +137,13 @@ def test_encoding(postgresql, data, execute):
     assert "blocking éléphant" in blocking.query
 
 
+def test_InvalidTextRepresentation(postgresql, data, execute):
+    """Test for issue #275"""
+    postgresql.execute("select '123' ~ '\\d+', 'Hello world!\n', pg_sleep(3)")
```
dlax, line 142: Can you explain how this query triggers the error?

If I'm running ` psql -c "select '123' ~ '\\d+', 'Hello world!\n', pg_sleep(3);"` on one side, and, aside:
```python
>>> import psycopg2
>>> conn = psycopg.connect()
>>> conn.autocommit = True
>>> cur.execute("select query from pg_stat_activity"); cur.fetchall()
<psycopg.Cursor [TUPLES_OK] [IDLE] (database=denis) at 0x7f2fa674ff60>
[('',), ('',), ("select '123' ~ '\\\\d+', 'Hello world!\\n', pg_sleep(3);",), ('select query from pg_stat_activity',), ('select query from pg_stat_activity',), ('',), ('',), ('',)]
```
I see no problem. Am I missing something?

blogh: It's a test that should have been included when we introduced the cast to bytea.

I'd like to add it to make sure the solution we choose covers this (I wonder 
what will append if we end up using `psycopg2.extensions.register_type(psycopg2.extensions.BYTES)` ).

I'll remove it later if it's useless.

PR #340, opened 2023-01-24
Prepare queries when using the Psycopg 3 database driver

dlax:
Based on #335


dlax:
@blogh, if you want to give it a try; otherwise, I'll merge by the end of the week.

PR #343, opened 2023-02-27
Retrieve database encoding

dlax:
This might be useful in forthcoming changes.


pgactivity/types.py
```diff
@@ -907,6 +907,7 @@ class BaseProcess:
     duration: Optional[float]
     state: str
     query: Optional[str]
+    encoding: str = attr.ib(converter=converters.default_if_none("utf-8"))  # type: ignore[misc]
```
blogh, line 910: The default value should be `server_encoding`, which is usually `utf-8`.
Not sure it's worth changing.

blogh (approved):

pgactivity/types.py
dlax: Indeed. I'll try to use that data later on (it's not easy now as we don't have the connection available at this point).

Issue #348, opened 2023-03-07
3.1.1: TypeError: cannot use a string pattern on a bytes-like object


esabol:
Hi. The `README.md` says Python 3.7 and above is supported? I'm using Python 3.8.5, but I'm getting the following error:
```
Traceback (most recent call last):
  File "/path/to/home/.local/bin/pg_activity", line 8, in <module>
    sys.exit(main())
  File "/path/to/home/.local/lib/python3.8/site-packages/pgactivity/cli.py", line 389, in main
    dataobj = data.pg_connect(
  File "/path/to/home/.local/lib/python3.8/site-packages/pgactivity/data.py", line 518, in pg_connect
    data = Data.pg_connect(
  File "/path/to/home/.local/lib/python3.8/site-packages/pgactivity/data.py", line 107, in pg_connect
    pg_version = pg_get_short_version(pg_get_version(pg_conn))
  File "/path/to/home/.local/lib/python3.8/site-packages/pgactivity/data.py", line 57, in pg_get_short_version
    res = re.match(
  File "/usr1/local/anaconda_py3/miniconda38/lib/python3.8/re.py", line 191, in match
    return _compile(pattern, flags).match(string)
TypeError: cannot use a string pattern on a bytes-like object

> ~/.local/bin/pg_activity --version
pg_activity 3.1.1
> /usr1/local/anaconda_py3/miniconda38/bin/python3 --version
Python 3.8.5
```
I installed it using the command `python3 -m pip install "pg_activity[psycopg]" --user`.

Thanks,
Ed

dlax:
That's strange, I cannot reproduce (tried to install miniconda, which is Python 3.8.16 at the moment).

Can you try to run this python session from `/usr1/local/anaconda_py3/miniconda38/bin/python3` and report back?
```python
>>> from pgactivity import pg
>>> pg.__version__
>>> conn = pg.connect()
>>> pg.fetchone(conn, "select version()")
```

dlax:
Ok; then can you try?
```python
>>> import psycopg
>>> conn = psycopg.connect("<your connection string>")
>>> conn.execute("select version()").fetchone()
```

esabol:
`(b'PostgreSQL 10.23 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-18), 64-bit',)`

esabol:
Using `.decode()` to convert the bytes to a string works just fine, also:
```
[...]

PR #350, opened 2023-03-09
improve handling of client encoding

dlax:
Resolves part of #348, more specifically, the original problem statement not the one from https://github.com/dalibo/pg_activity/issues/348#issuecomment-1461039555 which will be addressed separately.


blogh (approved):
LGTM

PR #351, opened 2023-03-09
convert postgres encoding to python ones

dlax:
Addresses second part of #348.

### DIFF
--- a/pgactivity/pg.py
+++ b/pgactivity/pg.py
@@ -20,10 +20,12 @@ try:
 
     import psycopg
     from psycopg import sql as sql
+    from psycopg._encodings import pg2pyenc
     from psycopg.adapt import Buffer, Loader
     from psycopg.conninfo import make_conninfo, conninfo_to_dict
     from psycopg.rows import dict_row
     from psycopg.errors import (
+        NotSupportedError,
         FeatureNotSupported as FeatureNotSupported,
         InterfaceError as InterfaceError,
         InvalidPassword as InvalidPassword,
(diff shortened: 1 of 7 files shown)


pgactivity/pg.py
```diff
@@ -169,6 +171,16 @@ def fetchall(
         with cursor(conn, mkrow, text_as_bytes) as cur:
             return cur.execute(query, args, prepare=True).fetchall()
 
+    def decode(value: bytes, pgenc: bytes, *, errors: str) -> str:
+        """Decode 'value' with PostgreSQL encoding 'pgenc' converted to Python encoding
+        name if available.
+        """
+        try:
+            pyenc = pg2pyenc(pgenc)
+        except NotSupportedError:
```
blogh, line 180: This case is for psycopg2 ?

pgactivity/pg.py
```diff
@@ -246,6 +258,18 @@ def fetchall(  # type: ignore[no-redef]
             return [mkrow(**row) for row in rows]
         return rows
 
+    def decode(value: bytes, pgenc: bytes, *, errors: str) -> str:
+        """Decode 'value' with PostgreSQL encoding 'pgenc' converted to Python encoding
+        name if available.
+        """
+        import codecs
```
blogh, line 265: Is there a reason you import here ?

blogh (approved):

pgactivity/pg.py
dlax: Not sure I get your question... This function is for psycopg 3+, there is a similar function defined for psycopg2 below. If you ask about the try/except, no it's not for psycopg2 either, it is because some postgres encodings do not have an equivalent in Python and we thus fall back to utf-8; typically that's the case of `EUC_TW` which is covered in added tests.

pgactivity/pg.py
dlax: I'll move it up.

PR #352, opened 2023-03-10
Preserve PGCLIENTENCODING before setting client_encoding=auto

dlax:
Follow-up d91a36487d17df3997b6fd202ae4d2917dded688

### DIFF
--- a/pgactivity/pg.py
+++ b/pgactivity/pg.py
@@ -50,11 +50,12 @@ try:
             return data.decode(errors="replace")
 
     def connect(dsn: str, **kwargs: Any) -> Connection:
-        # Set client_encoding to 'auto', if unset by the user.
-        # This is (more or less) what's done by psql.
-        conninfo = conninfo_to_dict(dsn)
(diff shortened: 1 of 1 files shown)


pgactivity/pg.py
```diff
@@ -50,11 +50,12 @@ def load(self, data: Buffer) -> str:
             return data.decode(errors="replace")
 
     def connect(dsn: str, **kwargs: Any) -> Connection:
-        # Set client_encoding to 'auto', if unset by the user.
-        # This is (more or less) what's done by psql.
-        conninfo = conninfo_to_dict(dsn)
-        conninfo.setdefault("client_encoding", "auto")
-        dsn = make_conninfo(**conninfo)
+        if "PGCLIENTENCODING" not in os.environ:
+            # Set client_encoding to 'auto', if unset by the user.
```
blogh, line 54: ``
```suggestion
            # Set client_encoding to 'auto', if it's not set.
```

On most installation I've seen, it's not set.

blogh (approved):
Aside from the comment I made LGTM.

PR #357, opened 2023-03-16
Delimit header columns with a comma instead of a dot

### LINKED ISSUE #230: Invalid character on interface
Apparently the PIPE was replaced by an unusual character, breaking the interface.
Please go back to PIPE. Works great and doesn't break the screen

image

### LINKED ISSUE #356: Weird display issues inside GNU screen
If I start pg_activity inside of an xterm window, I see this:
Screen Shot 2023-03-15 at 3 53 48 PM
Looks very nice! I think this is how it is supposed to look, although there are no blue dots/bullets between the items at the top. That's probably just be a font issue though. $TERM is "xterm256-color".

If I start GNU screen inside of the same xterm window, I see this:
Screen Shot 2023-03-15 at 4 01 05 PM
This doesn't look right. All the items at the top are on separate lines for some reason? And some lines are completely missing, like the database version and server name.

P.S. I figured out why I was getting `â` instead of the bullets/dots in my macOS Terminal window. Terminal profile was set to Latin-1 instead of UTF-8. When I changed it to UTF-8, I got the blue bullets/dots. That said, can you point me to where in the code the blue dots/bullets are printed? I'd like to change that character to something else....

EDIT: Hmmm, using xterm on  RHEL 8.7 with $TERM set to "xterm-color" (not inside of GNU screen), I get blue squares instead of the blue dots/bullets:
Screen Shot 2023-03-15 at 4 32 59 PM

dlax:
The character we were previously using, U+22C5 DOT OPERATOR, is apparently not rendering well for everybody. Let's use a plain ASCII character instead.

Fix #356
Fix #230

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change log
 
+## UNRELEASED
+
+### Fixed
+
+* Replace the previous header column separator (`⋅`) by a comma to improve user
+  experience in situations where that character did not render well (#356,
+  #230).
+
 ## pg\_activity 3.2.0 - 2023-03-06
 
 ### Fixed
--- a/pgactivity/views.py
+++ b/pgactivity/views.py
@@ -188,7 +188,7 @@ def header(
         return f"{term.bold_green(hbytes)} - {term.bold_green(counts)}"
 
     def render_columns(
-        columns: Sequence[List[str]], *, delimiter: str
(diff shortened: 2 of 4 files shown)


esabol:
This fixes issue #356 in my testing. Thank you!

blogh (approved):

PR #362, opened 2023-03-23
Drop 'is_superuser' query file

dlax:
This is unused since commit 460826d6edc9a112f98980a1fd1ca494b04b52e0.


blogh (approved):
good catch, thx. lgtm

PR #366, opened 2023-03-28
improve rendering of 'client' column with IPv6 addresses

### LINKED ISSUE #365: The 'client' column only shows the first 16 characters of IPv6 addresses.
When working with IPv6 connections, the "client" column is only showing the first 16 characters of the client address, which is way too small to identify the client. [Here is the column definition with the fix to display the full IPv6 address](https://github.com/dalibo/pg_activity/blob/25aa3fca8a05d943809f84483f6ef84c6547017b/pgactivity/types.py#L328).

Here is one example:
Screenshot 2023-03-27 at 14 48 26

I don't have an ultimate solution for this, but a config flag (like --full-client-ip) to show the full client address (opt-in) might be doable and don't show the pretty long addresses by default.

dlax:
Fix #365

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+### Fixed
+
+* Avoid truncating IPv6 addresses in the `client` column.
+
 ### Added
 
 * Possibly shorten IPv6 addresses in the `client` column.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change log
 
+## UNRELEASED
+
+### Added
+
+* Possibly shorten IPv6 addresses in the `client` column.
+
 ## pg\_activity 3.3.0 - 2023-03-22
 
 ### Fixed
--- a/pgactivity/types.py
+++ b/pgactivity/types.py
@@ -326,7 +326,8 @@ class UI:
                 key="client",
                 name="CLIENT",
                 min_width=16,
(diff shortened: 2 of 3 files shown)


brainexe:
Wow, that was fast, full ipv6 are shown now!

but as note: the used `client.client_addr::TEXT` is something like `2b00:2f78:fffd:1234::567/128` or `35.236.54.21/32`, which is always throwing an Exception in the new formatting function like `XXX does not appear to be an IPv4...` as it contains the network mask and always just returning the given raw string.

instead of `activity.client_addr::TEXT` a `host(activity.client_addr)` might be more suitable in the queries to get the client IP address ([looks like also available for old PG versions](https://www.postgresql.org/docs/8.3/functions-net.html#CIDR-INET-FUNCTIONS-TABLE))  

Right now:
Screenshot 2023-03-28 at 15 51 46

brainexe:
Oh just noticed: the header is a bit off:
Screenshot 2023-03-29 at 08 12 01

As a quick test I replaced the `a.client_addr::TEXT` by `'9b05:2d54:9be9:d1b6:385b:5124:b6ee:9aa9/128'` within the queries.

dlax:
> Oh just noticed: the header is a bit off with full ipv6 ones:

Oops, that's more problematic in fact. I need to think more about this. In the meantime, we can move forward with #368 and I'll rework this PR based on that.

dlax:
Superseded by #373

PR #368, opened 2023-03-29
Preserve the 'inet' type of client_addr column

dlax:
Instead of casting the client_addr to TEXT in PostgreSQL we keep the inet value which will be handled fine by psycopg. This brings us a nicer representation of some values, e.g. from 192.168.1.5/32 previously to 192.168.1.5. This will also help for IPv6 addresses which, if shortened that way, might fit better within the max width of the "client" column.

This is an alternative solution towards fixing #365, but not a complete solution as the column width is still limited to 16.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## UNRELEASED
 
+### Added
+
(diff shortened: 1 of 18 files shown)


blogh (approved):
Nice.

PR #369, opened 2023-04-07
Add a --width=COLUMN:VALUE option

### LINKED ISSUE #365: The 'client' column only shows the first 16 characters of IPv6 addresses.

dlax:
Fix #365

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## UNRELEASED
 
+### Added
+
+* Improve rendering of the `client` column by possibly abbreviating IP
+  addresses.
+
 ### Fixed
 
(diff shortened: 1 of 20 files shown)


dlax:
Replaced by #370

PR #370, opened 2023-04-11
Add a --width=COLUMN:VALUE option

dlax:
Fix #366

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@
 
 * Fix a crash when trying to reconnect.
 
+### Added
+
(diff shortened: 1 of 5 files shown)


dlax:
Thinking more about this, I wonder if this wouldn't be better if handled through a configuration file. We have none at the moment, but the idea was raised in the past already, so maybe it'd be time to introduce this?

blogh:
I like the idea or a config file for such things.

dlax:
> I like the idea of a config file for such things.

Ok; thank you. Let's move foward with #373 then.

PR #373, opened 2023-04-19
add a configuration to control columns visibility and width

dlax:
This is an alternative to #370 which adds support for configuring pg_activity from a file. ~~The JSON format is to be discussed.~~ (EDIT: changed to INI)

What do you think @blogh?

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,23 @@ ex:
       --no-proc-info        Display workers process information in header.
       --refresh REFRESH     Refresh rate. Values: 0.5, 1, 2, 3, 4, 5 (default: 2).
 
+## Configuration
+
+`pg_activity` may be configured through configuration file, in [INI format][],
+read from `${XDG_CONFIG_HOME:~/.config}/pg_activity.conf` or
+`/etc/pg_activity.conf` in that order. Command-line options may override
+configuration file settings.
+This is used to control how columns in the processes table are rendered, e.g.:
+```ini
+[client]
+hidden = yes
+
+[database]
+width = 9
+```
+
+[INI format]: https://docs.python.org/3/library/configparser.html#supported-ini-file-structure
+
 ## Notes
 
 Length of SQL query text that `pg_activity` reports relies on PostgreSQL
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@
 
 * Improve rendering of the `client` column by possibly abbreviating IP
   addresses.
+* Add support for configuring pg\_activity from a configuration file in INI
+  format and installed at `${XDG_CONFIG_HOME:~/.config}/pg_activity.conf` or
+  `/etc/pg_activity.conf`.
+* Determine which columns to hide from reading the configuration file (`hidden =
+  yes|no` option).
+* Determine columns width from the configuration file (`width = <positive
+  integer>` option).
 
 ### Fixed
 
--- a/pgactivity/cli.py
+++ b/pgactivity/cli.py
@@ -10,6 +10,7 @@ from typing import Optional
 from blessed import Terminal
 
 from . import __version__, data, types, ui
+from .config import Configuration, ConfigurationError
 from .pg import OperationalError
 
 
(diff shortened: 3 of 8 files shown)


dlax:
> The JSON format is to be discussed.

We can also use an INI format, if we anticipate that we'll never need nested options (or sub-sections).

blogh:
At first, I was not fond of the idea of  a `json` file, but it's not that bad. I think we can keep this format.

Do we remove these options from the cli ?

```
Process table display options:
  These options may be used hide some columns from the processes table.

  --no-pid              Disable PID.
  --no-database         Disable DATABASE.
  --no-user             Disable USER.
  --no-client           Disable CLIENT.
  --no-cpu              Disable CPU%.
  --no-mem              Disable MEM%.
  --no-read             Disable READ/s.
  --no-write            Disable WRITE/s.
  --no-time             Disable TIME+.
  --no-wait             Disable W.
  --no-app-name         Disable App.
```

dlax:
> At first, I was not fond of the idea of a |json| file, but it's not that 
> bad. I think we can keep this format.

Not very happy with JSON either; it's just the simplest solution.

But will think about other options and try to see how it works.

> Do we remove these options from the cli ?

I don't think it's needed for a breaking change. As is, command-line 
options take precedence over the configuration file so this is properly 
handled. Also, from a user point of view, I suspect people would want to 
sometime hide a specific column one-shot through the --no-xxx option, 
but probably won't change columns width so often.

blogh:
That makes sens, yes.

dlax:
Changed to INI format; now ready for testing and review @blogh.

README.md
```diff
@@ -138,6 +138,23 @@ ex:
       --no-proc-info        Display workers process information in header.
       --refresh REFRESH     Refresh rate. Values: 0.5, 1, 2, 3, 4, 5 (default: 2).
 
+## Configuration
+
+`pg_activity` may be configured through configuration file, in [INI format][],
+read from `${XDG_CONFIG_HOME:~/.config}/pg_activity.conf` or
+`/etc/pg_activity.conf` in that order. Command-line options may override
```
blogh, line 145: Should we specify that it's exclusive ?

blogh (changes requested):
How would it work if we add a panel with different columns (I suppose it would work, we already have mem/cpu/read/write)?

Does it make sense to hide columns in some panels but not in others, should we support it? (eg displaying client only for blocking queries)

README.md
dlax: It's not "exclusive"; rather command-line options take precedence.

dlax:
> How would it work if we add a panel with different columns (I suppose it 
> would we already have mem/cpu/read/write)?

It works the same as before: if a column is configured to be hidden, it 
will be hidden in all panels.

> Does it make sense to hide columns in some panels but not in others? (eg 
> displaying client only for blocking queries)

I don't know, maybe... but maybe not...
Yet, if we want a way to configure per-panel visibility, we'd need a way 
to select the panel name from the configuration file. Since there is no 
possibility to nest options in INI format, the nesting should be in the 
section mame, i.e. something like `[running.client]`. Or we should 
rather go with a file format supporting nested data structures: but is 
it worth?

README.md
blogh: I mean it's `/etc/pg_activity.conf` OR `~user/.config/pg_activity.conf`.

dlax: That's how it's written already: "read from `${XDG_CONFIG_HOME:~/.config}/pg_activity.conf` or `/etc/pg_activity.conf`". Seems explicit enough to me.

dlax:
I'm in favor of the INI format. It's certainly simpler for the user and
the complexity on our side, arguably a bit more important than JSON (but
without the need for a third-party dependency), is acceptable.

README.md
blogh: I first understood it as either. But it's probably my English acting up.

blogh:
Ok, let commit to this then. Thanks !

PR #385, opened 2023-07-01
Add a new output column "rollback ratio"

kmoppel-cognite:
To the global / instance info section. To quickly understand if something wrong with queries

Screenshot from 2023-07-01 11-36-17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change log
 
+## Unreleased
+
+### Added
+
+* The *rollback ration* is now displayed in the "global" header (#385).
+
 ## pg\_activity 3.4.2 - 2023-06-01
 
 ### Fixed
--- a/pgactivity/data.py
+++ b/pgactivity/data.py
@@ -349,6 +349,7 @@ class Data:
         replication_slots = self.pg_get_replication_slots()
 
         hr: Optional[Pct] = None
+        rr: Optional[Pct] = None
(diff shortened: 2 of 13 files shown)


kmoppel-cognite:
This makes that view row a bit more longer than others - not a real problem for most screens I guess, but would a bit nicer if things are more aligned. Ideas to "compact" the existing columns:
1) Shorten the "uptime" when its in > 1day - no real need for "minutes" then
2) Remove "and" from "uptime" - unnecessary, the shorter the better
3) "cache hit ratio" -> "cache ratio"

Does that sound like an OK idea ? Could send another PR for that

dlax (approved):
This looks good to me; thank you @kmoppel-cognite!

@blogh, what do you think?

* * *
TODO before merging:
- [x] squash commits
- [x] add an entry in the changelog

blogh:
Thanks for the addition.

As fat as compacting the columns go : it's a good idea too.

I would go for "cache hit %" instead of "cache ratio". 
Same for "rollback %".

blogh (approved):

dlax:
> TODO before merging:

I'll take care of this now, and then merge.

dlax:
Merged; thank you @kmoppel-cognite.

dlax:
@kmoppel-cognite:
> Ideas to "compact" the existing columns:
> [...]
> Does that sound like an OK idea ? Could send another PR for that

Sounds good; please go ahead.

PR #387, opened 2023-07-02
Add new --queries-only flag for best viewport usage and performance

kmoppel-cognite:
As there are quite many viewport customization flags, it's currently quite tedious to set things for most compact viewport usage, when on some terminal withing a terminal, as it quite often is on the clouds.

Other effect is that in this mode we're a bit lighter on the monitored DB, skipping some queries.

### DIFF
--- a/docs/man/pg_activity.1
+++ b/docs/man/pg_activity.1
@@ -507,6 +507,11 @@ required by another session. It shows following information:
 .Vb 1
 \&        Display workers process information in header.
 .Ve
+.IP "\fB\-\-no\-inst\-info\fR" 2
+.IX Item "--queries-only"
+.Vb 1
+\&        Show only PID, user, time, waiting and query columns without any information headers.
+.Ve
 .IP "\fB\-\-refresh\fR" 2
 .IX Item "--refresh"
 .Vb 1
--- a/pgactivity/cli.py
+++ b/pgactivity/cli.py
@@ -3,7 +3,7 @@ import os
 import socket
 import sys
 import time
-from argparse import ArgumentParser
+from argparse import ArgumentParser, Namespace
 from io import StringIO
 from typing import Optional
 
@@ -347,6 +347,14 @@ def get_parser() -> ArgumentParser:
         help="Display workers process information in header.",
         default=True,
     )
+    # --queries-only
+    group.add_argument(
+        "--queries-only",
+        dest="queries_only",
+        action="store_true",
+        help="Show only PID, user, time, waiting and query columns without any information headers.",
+        default=False,
+    )
     # --refresh
     group.add_argument(
         "--refresh",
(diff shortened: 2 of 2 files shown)


kmoppel-cognite:
Screenshot from 2023-07-02 11-05-31

dlax:
I'm not sure adding another CLI option is the way to go because it seems to me that this one is just a shortcut for a bunch of `--no-xxx` options. Also I believe the selection of things to hide/show is a matter of taste.

On the other hand, recently, we introduced support for tweaking the UI through a [configuration file](https://github.com/dalibo/pg_activity/#configuration). At the moment, configuration options are limited to columns (setting their width and possibly hiding them) and currently, part of what you want can be achieved by:
```
$ cat ~/.config/pg_activity.conf
[database]
hidden = true

[client]
hidden = true

[cpu]
hidden = true

[...]
```
Yet it'd be totally fine to extend the configuration file so as to allow hiding other things as you suggest here.

What that suit your needs @kmoppel-cognite?

kmoppel-cognite:
@dlax Yes, there are quite some flags for sure, I've been using pg_activity almost every working day for the past 7 or 8 years :) But at the same time I was finally annoyed enough to do a PR as right now it takes almost a minute to set all those flags, when including looking at the help.

The config file option is indeed very nice for a more classical workflow, where you have one or few central monitoring nodes - but in my case for example where we have dozen+ different cloud projects with hundreds of DBs, where CLI interactions happen on ephemeral K8s pods, so that for ad-hoc troubleshooting I constantly need to start new pods and set all those 8 or 9 flags to blend out extra noise.

Maybe if to just name it better to `--minimal` so that it indeed would stay the only flag of this class? Also on 2nd though I'd now leave out the `--no-inst-info` flag to show global session activity. It could be disabled additionally to `--minimal` if needed.

Wdyt?

dlax:
Thank you for clarifying your use cases @kmoppel-cognite; now it seems more reasonable to me.

I like the idea of this "minimal" UI that could be enabled through a simple CLI flag. But rather than adding such a flag, I wonder if we could implement some kind of "configuration profiles", e.g. by loading them from `~/.config/pg_activity/<profile-name>.conf` upon `pg_activity --profile <profile-name>` invocation. In addition, we could provide a set of *built-in* profiles, such as the "minimal" one you're suggesting, that would be available without any configuration file. That would mean that invoking `pg_activity --profile minimal` would load the built-in "minimal" profile if no `minimal.conf` file exists or use the latter if it exists, as a mean to override a built-in profile.

(also cc @blogh)

blogh:
I like the profile idea.

dlax:
@kmoppel-cognite, I implemented the profile idea discussed above in #395. Would you like to take a look and see if it fits your needs?

dlax:
@kmoppel-cognite, I finally found the time to work on the profile idea. The master branch already supports user profiles (see https://github.com/dalibo/pg_activity/?tab=readme-ov-file#configuration) and https://github.com/dalibo/pg_activity/pull/403 suggests to add built-in profiles. Can you have a look at these changes and tell us if this suits your needs?
Thank you.

dlax:
Now that #403 (and other preparatory works) have been merged, I am closing this PR. @kmoppel, feel free to share your feedback on these improvements here though.

PR #395, opened 2023-08-23
Add support for configuration profiles

dlax:
This stems from the idea discussed at https://github.com/dalibo/pg_activity/pull/387#issuecomment-1623178466.#396.

```
$ cat ~/.config/pg_activity/my.conf 
[database]
hidden = true
```
```
$ pg_activity --profile my  # will use the file above as configuration
```

One thing I wonder is whether the option should be named `--config <full-path-to-config>` or if we should introduce another option later on.

* * *

Based on #404

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - python: "3.7"
+          - python: "3.8"
             psycopg: "psycopg2"
           - python: "3.12"
             psycopg: "psycopg3"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,11 @@ repos:
         exclude: tests/test_*.txt
   - repo: local
     hooks:
+      - id: pyupgrade
+        name: pyupgrade
+        entry: pyupgrade --py38-plus --exit-zero-even-if-changed
+        language: system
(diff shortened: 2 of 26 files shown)


blogh:
From what I understand the configuration we can set in the file relate to the columns of the query display. 
@kmoppel-cognite was looking for a way to hide sections of the header (instance, system, workers).

```
  --no-inst-info        Display instance information in header.
  --no-sys-info         Display system information in header.
  --no-proc-info        Display workers process information in header.
```

But I guess other options would be nice to have to.

Did I miss something ?

dlax:
>  From what I understand the configuration we can set in the file relate to the columns of the query display.
> @kmoppel-cognite was looking for a way to hide sections of the header (instance, system, workers).

These are not configurable through configuration files at the moment, 
indeed; but we can certainly add support for these later on.

kmoppel:
Hey, glad to see some progress! Looks promising :)  At the same time - do I understand correctly that currently there is still no "infra" for built-in profiles where configs could be read in some fallback mode from some "flags dictionary" in code?

dlax:
> do I understand correctly that currently there is still no "infra" for built-in profiles where configs could be read in some fallback mode from some "flags dictionary" in code?

Sorry, I don't get your question; can you reformulate?

blogh (approved):
LGTM, thanks for adding this !

PR #403, opened 2024-02-19
add builtin profiles

dlax:
implements the idea suggested in https://github.com/dalibo/pg_activity/pull/387#issuecomment-1623178466


dlax:
The content of built-in profiles is to be discussed. More profiles may be added, also. Suggestions welcome.

blogh (approved):
LGTM, great idea !

PR #404, opened 2024-02-20
make it possible to control header items display from configuration file

dlax:
based on #405 and #406; only last tree commits are new.


blogh (approved):
LGTM

PR #406, opened 2024-02-21
Upgrade type annotations syntax

dlax:
Based on #405


blogh (approved):
Ok, Interesting read thx.

PR #408, opened 2024-03-22
make the default color of cells configurable

### LINKED ISSUE #407: columns not visible with terminal in GNOME dark or similar
When using pg_activity with terminal in dark mode, the column's database, app and user are not visible, they use "black_bold" as color and it's not visible.

Could be an option to have two different color profiles? One for dark mode and one for light mode?

image

The workaround is modifying colors.py and change black_bold to white_bold in this columns

dlax:
Fix #407


blogh (approved):
LGTM

PR #416, opened 2024-04-09
fix configuration of appname width

### LINKED ISSUE #415: appname width config option does not work
I've noticed, that `width` config option in `/etc/pg_activity.conf` does not work for `[appname]` column, while it does work for other columns. It is strange, given, that `hidden` option does work for the same column.

Contents of config file `/etc/pg_activity.conf`:
```ini
[appname]
width = 1
```
Command line options: `pg_activity -p 5432`

There is no user-specific config file nor profile (directory `~/.config/pg_activity` does not exist).

dlax:
Fix #415


blogh (approved):
LGTM

PR #420, opened 2024-05-28
Add a 'y' command to copy focused query to clipboard

### LINKED ISSUE #311: Add a shortcut to copy a query
I'd like to easily and quickly grab a copy of a query that's been running for a long time in order to save it off before killing it. Even better if it was a keyboard shortcut that copied to my system clipboard.

Perhaps [pyperclip](https://pypi.org/project/pyperclip/) can be of use?

dlax:
Fix #311


blogh (approved):
Tried it with the "contour" terminal. 
It worked.

Issue #425, opened 2024-08-20
feature request: xmin info


alexitheodore:
I use pg_activity all day every day, thanks!

One feature that feels missing though is information on xid. I struggle to keep my long-running queries at bay (running a datawarehouse type db) but sometimes they happen and cause issues. Knowing what transactions are blocking vacuum based on xmin horizon stuff would be really helpful. Its not always as simple as "the oldest transaction" because that could be vacuums or replication slots OR (in my case) procedures that have transaction control built in internally.

blogh:
It took me while but it's in.

PR #431, opened 2025-02-13
Add the xmin horizon to running query display

blogh:
The xmin horizon was added in 9.4 in commit [1], the addition of the column was asked in #425.

[1] https://github.com/postgres/postgres/commit/dd1a3bccca241a998b4ebf39d67202698e5fa599

### DIFF
--- a/docs/man/pg_activity.1
+++ b/docs/man/pg_activity.1
@@ -182,6 +182,8 @@ The running queries panel shows all running queries, transactions or backends
 .IP "\- \fBPID\fR: process id of the backend which executes the query;" 2
 .IX Item "- PID: process id of the backend which executes the query;"
 .PD 0
+.IP "\- \fBXMIN\fR: xmin horizon of the backend;" 2
+.IX Item "- XMIN: xmin horizon of the backend;"
 .IP "\- \fBDATABASE\fR: database specified in the connection string;" 2
 .IX Item "- DATABASE: database specified in the connection string;"
 .IP "\- \fBAPP\fR: application name specified in the connection string;" 2
@@ -363,6 +365,11 @@ required by another session. It shows following information:
 .IX Item "--pid, --no-pid"
 .Vb 1
 \&        Enable/disable PID.
+.Ve
+.IP "\fB\-\-xmin\fR, \fB\-\-no\-xmin\fR" 2
+.IX Item "--xmin, --no-xmin"
+.Vb 1
+\&        Enable/disable XMIN.
 .Ve
 .IP "\fB\-\-database\fR, \fB\-\-no\-database\fR" 2
 .IX Item "--database, --no-database"
--- a/tests/data/local-processes-input.json
+++ b/tests/data/local-processes-input.json
@@ -2,6 +2,7 @@
     "fs_blocksize": 4096,
     "new_processes": {
         "6221": {
+            "xmin": 1,
             "application_name": "pgbench",
             "client": null,
             "cpu": null,
@@ -55,6 +56,7 @@
             "write": null
         },
         "6222": {
+            "xmin": 2,
             "application_name": "pgbench",
             "client": null,
             "cpu": null,
@@ -108,6 +110,7 @@
             "write": null
         },
         "6223": {
+            "xmin": 3,
             "application_name": "pgbench",
             "client": null,
             "cpu": null,
@@ -161,6 +164,7 @@
             "write": null
         },
         "6224": {
+            "xmin": 4,
             "application_name": "pgbench",
             "client": null,
             "cpu": null,
@@ -214,6 +218,7 @@
             "write": null
         },
         "6225": {
+            "xmin": 5,
             "application_name": "pgbench",
             "client": null,
             "cpu": null,
@@ -267,6 +272,7 @@
             "write": null
         },
         "6226": {
+            "xmin": 6,
             "application_name": "pgbench",
             "client": null,
             "cpu": null,
@@ -320,6 +326,7 @@
             "write": null
         },
         "6227": {
+            "xmin": 7,
             "application_name": "pgbench",
             "client": null,
             "cpu": null,
@@ -373,6 +380,7 @@
             "write": null
         },
         "6228": {
+            "xmin": 8,
             "application_name": "pgbench",
             "client": null,
             "cpu": null,
@@ -426,6 +434,7 @@
             "write": null
         },
         "6229": {
+            "xmin": 9,
             "application_name": "pgbench",
             "client": null,
             "cpu": null,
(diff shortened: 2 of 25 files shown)


docs/man/pg_activity.1
```diff
@@ -1,4 +1,5 @@
-.\" Automatically generated by Pod::Man 4.14 (Pod::Simple 3.43)
+.\" -*- mode: troff; coding: utf-8 -*-
+.\" Automatically generated by Pod::Man v6.0.2 (Pod::Simple 3.45)
```
dlax, line 2: Can you make a separate commit about the change of `pod2man` version alone, without changing `.pod` file, and then commit the latter (and regenerate `.1`) along with the new feature?

pgactivity/views.py
```diff
@@ -12,7 +12,10 @@
 from . import colors, utils
 from .activities import sorted as sorted_processes
 from .compat import link
-from .keys import BINDINGS, EXIT_KEY
+from .keys import (
+    BINDINGS,
+    EXIT_KEY,
+)
```
dlax, line 18: This hunk is unrelated, can you revert it?

tests/data/local-processes-input.json
```diff
@@ -2,6 +2,7 @@
     "fs_blocksize": 4096,
     "new_processes": {
         "6221": {
+            "xmin": "1",
```
dlax, line 5: Should be an integer, as declared in `RunningProcess`, no?

tests/test_views.txt
```diff
@@ -212,6 +212,7 @@ Tests for processes_rows()
 >>> processes1 = [
 ...     LocalRunningProcess(
 ...         pid="6239",
+...         xmin="1",
```
dlax, line 215: Same here, we could use integers, no? (The same is probably true for `pid`, but that's unrelated and so cannot be changed now.)

pgactivity/ui.py
```diff
@@ -156,7 +156,7 @@ def main(
                 else:
                     pg_procs.reset()
                     ui.end_interactive()
-                    changes = {
+                    changes: types.Changes = {
```
dlax, line 159: This change is also unrelated to the new feature, but it's probably due to a new Mypy version (more strict) which we didn't test before (I'm seeing the error as well on master).

In fact, I'd simply suggest:
```suggestion
                    changes: dict[str, Any] = {
```

If you can commit this in a distinct commit, that's also be better.

dlax (changes requested):
Looks good; just a couple of minor comments.

Also, I'm wondering if you put the new column close to left side intentionally? Is there a business logic reason for this?

pgactivity/views.py
blogh: It was requested by isort.

pgactivity/ui.py
blogh: Yes it's due to mypy.

pgactivity/views.py
blogh:
```
pg_activity on  xmin_425 [$?⇕] via 🐍 v3.13.2 (.venv)
❯ isort ./pgactivity/
Fixing /home/benoit/git/dalibo/pg_activity/pgactivity/views.py

pg_activity on  xmin_425 [$!?⇕] via 🐍 v3.13.2 (.venv)
❯ git diff
diff --git a/pgactivity/views.py b/pgactivity/views.py
index b8720af..190905a 100644
--- a/pgactivity/views.py
+++ b/pgactivity/views.py
@@ -12,7 +12,10 @@ from blessed import Terminal
 from . import colors, utils
 from .activities import sorted as sorted_processes
 from .compat import link
-from .keys import BINDINGS, EXIT_KEY
+from .keys import (
+    BINDINGS,
+    EXIT_KEY,
+)
 from .keys import HELP as HELP_KEY
 from .keys import (
     KEYS_BY_QUERYMODE,
```

pgactivity/ui.py
blogh:
```
pg_activity on  xmin_425 [$!?⇕] via 🐍 v3.13.2 (.venv)
❯ mypy ./pgactivity/
pgactivity/ui.py:164: error: Incompatible types in assignment (expression has type "float", target has type "int")  [assignment]
pgactivity/ui.py:169: error: Incompatible types in assignment (expression has type "QueryMode", target has type "int")  [assignment]
pgactivity/ui.py:174: error: Incompatible types in assignment (expression has type "SortKey", target has type "int")  [assignment]
Found 3 errors in 1 file (checked 17 source files)
```

pgactivity/views.py
blogh: It's in a separate commit.

pgactivity/ui.py
blogh: It's in a separate commit.

blogh:
The tests crash on this although the read / write parts where not changed in this patch (`tests/test_views.txt`).

```
    -6239   1        pgbench                   pgbench         postgres            local 0.1    1.0  7B       12B       0.000000                N N        idle in trans UPDATE pgbench_accounts SET abalance = abalance + 141 WHERE aid = 1932841;
    +6239   1        pgbench                   pgbench         postgres            local 0.1    1.0  7B       12.3B     0.000000                N N        idle in trans UPDATE pgbench_accounts SET abalance = abalance + 141 WHERE aid = 1932841;
    -6228   2        pgbench                   pgbench         postgres            local 0.2    1.0  0B       1.08M     0.000413                  Y               active \_ UPDATE pgbench_accounts SET abalance = abalance + 141 WHERE aid = 1932841;
    +6228   2        pgbench                   pgbench         postgres            local 0.2    1.0  0.2B     1.08M     0.000413                  Y               active \_ UPDATE pgbench_accounts SET abalance = abalance + 141 WHERE aid = 1932841;
```

I also don't get this crash in my local env... 
Do you have an idea of why the display of these floats would change ?

dlax:
> The tests crash on this although the read / write parts where not changed in this patch (`tests/test_views.txt`).

No idea why; I can take a closer look today or at least before the end of the week.

In the meantime, I have pushed your three other fixes (manpage, mypy, isort) to master (you might want to rebase the remaining commit here). The test failure reproduces there so it's indeed unrelated to this PR.

dlax:
It's due to humanize 4.12.0, I can reproduce locally after `pip install -U humanize`.

dlax:
Fixed in https://github.com/dalibo/pg_activity/pull/432.

blogh:
> Also, I'm wondering if you put the new column close to left side intentionally? Is there a business logic reason for this?

No, it was easier while I was testing and it seemed as good a place as any.

dlax (approved):

PR #432, opened 2025-02-18
Avoid humanize 4.12.0

dlax:
We are apparently affected by this issue:

  https://github.com/python-humanize/humanize/issues/239

Let's avoid that version and wait for a fix.


blogh (approved):
Awesome thx !

PR #433, opened 2025-04-14
Disable log_statements when --hide-queries-in-logs is used

blogh:
We only disable log_min_duration_statement and log_min_duration_sample. when --hide-queries-in-logs which leaves us with tracs from log_statements. This aptch fixes that.

Authored by Kouber Saparev

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change log
 
+## Unreleased 
+
+### Fixed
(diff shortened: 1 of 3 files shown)


dlax (commented):
LGTM; can you add an entry in changelog file?

dlax (approved):

kouber:
Cheers.

PR #460, opened 2026-05-18
Specify which Python versions are actually supported and tested

### LINKED ISSUE #458: Clarification on supported Python versions in README
Hello,

I received the following downstream report: https://bugs.gentoo.org/973814

The README currently states:
`pg_activity can be installed using pip on Python 3.8 or later`

Could you please clearly specify which Python versions are actually supported and tested?

This would help avoid the impression that the project has been tested against a broader range of Python versions than it actually has.

Thanks.

blogh:
Fix #458

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,12 +311,7 @@ this would use `CTRL+SHIFT+V` to paste from this clipboard.
 
 **Python version supported and tests**
 
-pg\_activity is confirmed to run with Python 3.8 up to 3.14.
-
-The CI currently use :
-
-* python3.9 with psychopg2;
-* python13 with psychopg3.
+pg\_activity's CI run for version 3.10 to 3.14 included with psychopg3.
 
 # Hacking
 
--- a/README.md
+++ b/README.md
@@ -310,7 +310,8 @@ this would use `CTRL+SHIFT+V` to paste from this clipboard.
 
 **Python version supported and tests**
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -309,6 +309,15 @@ this escape sequence and set the clipboard accordingly. If so, the copy even
 works across remote connections (SSH). In general, terminal emulators supporting
 this would use `CTRL+SHIFT+V` to paste from this clipboard.
 
+**Python version supported and tests**
+
+pg\_activity is confirmed to run with Python 3.8 up to 3.14.
```
l00ptr: 3.8 or 3.9 ? See below for information about the tested versions: 3.8 or 3.9.

blogh: We don't test all python versions where pg_activity works in CI.
Should we? Only for python's supported versions (3.10>3.14)?
What do we do in our other projects?

l00ptr: As discussed today, I think we should only test and talk about the supported Python 3 version. So, in my humble opinion, for testing, we have two solutions:

- Test all versions exhaustively, but this will be time-consuming and painful.
- Only test the minimum and maximum versions (3.10 and 3.14).

If it works with an older Python version, that would be good for users in that situation, but we cannot guarantee that.

blogh: I used your second proposition.

PR #462, opened 2026-06-25
Add a new --strip-comments flag

kmoppel:
To optionally remove any comments from displayed queries using the "sqlparse" library. CSV output is not affected.


kmoppel:
As long comments often can hide more important SQL fragments getting displayed

blogh:
Thanks for the PR. It's a great addition.

blogh (approved):

